PR #3, opened 2019-08-02
feat: privacy rules. Closes #1.

### LINKED ISSUE #1: Feature Request: Sync namespace with private
Hi! Your plugin looks promising. Could you add these rules to help us keep the namespace and `private` in sync?

* `"private"`property must be either `true` or `false`. (Rule name could be `explicitly-decide-private`)
* If `"private": true` but package name starts with `"@"`: <s>Namespaced package probably should be public!</s> My project had a policy that namespaced package must be public. (Rule name could be `namespaced-requires-public`)
* If `"private": false` but package name doesn't start with `"@"`: Packages must be published in a namespace! (Rule name could be `public-requires-namespaced`)

They should not be auto-fixable, since it's not eslint's job to decide which property is the mistaken one.

zetlen:
Add rules for `required-properties`, `namespaced-requires-public`, and `public-requires-namespaced`.

Proposed to close #1.

### DIFF
--- a/docs/rules/public-requires-namespaced.md
+++ b/docs/rules/public-requires-namespaced.md
@@ -39,7 +39,6 @@ Example of **correct** code for this rule:
 ```
 
 The following code is safe because it is non-namespaced and private.
-ignores it.
 
 ```json
 {
--- a/docs/rules/public-requires-namespaced.md
+++ b/docs/rules/public-requires-namespaced.md
@@ -1,0 +1,57 @@
+# Require that public packages be namespaced
+
+If a package is configured to be public, then it must be namespaced (that is,
+its `"name"` field begins with `@`). This rule is the converse of `namespaced-requires-public`.
+
+## Rule Details
+
+If you want to make sure to publish all your packages under a namespace, this rule will prevent any non-namespaced package from being published.
+
+Examples of **incorrect** code for this rule:
+
+```json
+{
+    "name": "my-package",
+    "private": false,
+    "version": "1.0.0"
+}
+```
+
+The following code is incorrect because NPM assumes a package is public
+unless `"private"` is explicitly declared `true`, so this would inadvertently
+publish a non-namespaced package.
+
+```json
+{
+    "name": "my-package",
+    "version": "1.0.0"
+}
+```
+
+Example of **correct** code for this rule:
+
+```json
+{
+    "name": "@my-scope/my-package",
+    "private": false,
+    "version": "1.0.0"
+}
+```
+
+The following code is safe because it is non-namespaced and private.
+ignores it.
+
+```json
+{
+    "name": "my-package",
+    "private": true,
+    "version": "1.0.0"
+}
+```
+
+### Options
+
+There are no options for this rule.
+
+This rule is **not autofixable**; you must decide whether to declare it to be
+private or to add a namespace.
--- a/docs/rules/require-properties.md
+++ b/docs/rules/require-properties.md
@@ -42,7 +42,7 @@ Examples of **correct** code for this rule:
     "publishConfig": {
         "access": "restricted",
         "registry": "https://alternate.registry"
-    }
+    },
     "version": "1.0.0"
 }
 ```
--- a/docs/rules/require-properties.md
+++ b/docs/rules/require-properties.md
@@ -1,0 +1,66 @@
+# Require that certain properties be declared
+
+Some package properties are optional, but this rule can require them to be
+present.
+
+## Rule Details
+
+This rule only requires that the properties be declared; it does not validate
+that they are set to a particular value.
+
+Examples of **incorrect** code for this rule:
+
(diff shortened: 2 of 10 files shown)


docs/rules/public-requires-namespaced.md
```diff
@@ -0,0 +1,57 @@
...
+```json
+{
+    "name": "@my-scope/my-package",
+    "private": false,
+    "version": "1.0.0"
+}
+```
+
+The following code is safe because it is non-namespaced and private.
+ignores it.
```
mk-pmb: Who/what "ignores it"?

docs/rules/require-properties.md
```diff
@@ -0,0 +1,66 @@
...
+Examples of **correct** code for this rule:
+
+```json
+{
+    "name": "my-package",
+    "private": true,
+    "publishConfig": {
+        "access": "restricted",
+        "registry": "https://alternate.registry"
+    }
```
mk-pmb: Missing `,`. Might there be a linter for JSON inside markdown?

zetlen: Oh, there might be, but I've never set one up, Good catch!

docs/rules/public-requires-namespaced.md
zetlen: Whoops, left a trailing line. Fixed!

mk-pmb:
Hi! I finally found the time to try it, but it seems I'm doing something wrong.

My test repo: https://github.com/mk-pmb/try-eslint-plugin-package-json-privacy-mgmt-js
Travis CI error log: https://travis-ci.org/mk-pmb/try-eslint-plugin-package-json-privacy-mgmt-js/jobs/571072367

What could be the problem?

mk-pmb:
Ok now with installing from the PR branch it seems to somewhat work, albeit `package-json/valid-package-def` doesn't like the branch dep.
https://travis-ci.org/mk-pmb/try-eslint-plugin-package-json-privacy-mgmt-js/jobs/571078638
How do I tell it to not apply regular JS style rules?

zetlen:
@mk-pmb You're running @mk-pmb You can use an `overrides` config to do that, but oddly enough, that doesn't happen with any preset I've tried but yours. If you remove your preset in that sample repo, it lints just fine.

zetlen:
@mk-pmb Are you still having trouble with configuration overrides? Are they specific to this config, or do they happen whenever you add .json files to the lint target glob?

mk-pmb:
I'm not sure I understand the question. In the other thread I wrote which rules conflict. If I [remove those rules](https://github.com/mk-pmb/try-eslint-plugin-package-json-privacy-mgmt-js/commit/730ad8b3fd7f059b96a1b281687207d565d76954), the JSON file [is accepted](https://travis-ci.org/mk-pmb/try-eslint-plugin-package-json-privacy-mgmt-js/jobs/582093861).

mk-pmb:
Without your plugin, [eslint fails](https://travis-ci.org/mk-pmb/try-eslint-plugin-package-json-privacy-mgmt-js/jobs/582115935) with `Parsing error: Unexpected token :"` because JSON objects aren't a valid script. At least not in node.js:

```text
$ <<<'{ "foo": 123 }' nodejs; echo node rv=$?
[stdin]:1
{ "foo": 123 }
       ^

SyntaxError: Unexpected token :
```

mk-pmb:
> Are you still using it?

Nope. I'm still waiting for a version that works for my test repo. Then, hopefully, I'll be able to use it with my exttended linter configs as well. :-)

> Can you give me all the relevant versions (node, npm, eslint)?

eslint v6.6.0, Node.js v12.13.0, npm v6.12.0, Ubuntu 16.04.6 LTS xenial,
kernel: 4.15.0-62-lowlatency, machine/CPU/platform: x86_64 x86_64 x86_64

privatenumber:
Any way I can help making this mergeable? Would love to use the `required-properties` rule.

lib/rules/require-properties.js
```diff
@@ -0,0 +1,56 @@
...
+    create(context) {
+        return {
+            'Program:exit': node => {
+                if (!isPackageJson(context.getFilename())) {
+                    return;
+                }
+                const sourceCode = context.getSourceCode();
+                const packageRoot = extractPackageObjectFromAST(node);
+                const pkg = JSON.parse(sourceCode.getText(packageRoot));
+                const required = context.options[0] || [];
```
DarrenPaulWright, line 40: Should the default value include "name" and "version", since they are required fields? https://docs.npmjs.com/creating-a-package-json-file.
Then this could be a recommended rule.

mk-pmb:
Is there still hope?

JoshuaKGoldberg:
😄 There is hope! https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1#issuecomment-1716807880

PR #11, opened 2020-12-11
Add local-dependency rule 

kendallgassner:
Basically I keep making a mistake where I case am installing a local dependency using "link" and I mess up the casing:

This:
```
  "devDependencies": {
         "some-package": "link:../folder",
   }

```

Should be 
```
  "devDependencies": {
         "some-package": "link:../Folder",
   }

```
Npm doesn't support link but yarn does. When the path is wrong yarn does not sub out the link when releasing.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -31,7 +31,8 @@ module.exports.configs = {
         rules: {
             'package-json/sort-collections': 'error',
             'package-json/order-properties': 'warn',
-            'package-json/valid-package-def': 'error'
+            'package-json/valid-package-def': 'error',
+            'package-json/local-dependency': 'error'
         }
     }
 };
--- a/lib/rules/local-dependency.js
+++ b/lib/rules/local-dependency.js
@@ -1,0 +1,91 @@
+/**
+ * @fileoverview Checks existence of local dependencies in the package.json
+ * @author Kendall Gassner
+ */
+'use strict';
+const fs = require('fs');
+const path = require('path');
+const {
+    isPackageJson,
+    extractPackageObjectFromAST
+} = require('../processors/PackageJsonProcessor');
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+function fileExistsWithCaseSync(filepath) {
+    var dir = path.dirname(filepath);
+
+    if (dir === path.dirname(dir)) {
+        return true;
+    }
+    var filenames = fs.readdirSync(dir);
+    if (filenames.indexOf(path.basename(filepath)) === -1) {
+        return false;
+    }
+    return fileExistsWithCaseSync(dir);
+}
+
+module.exports = {
+    meta: {
+        docs: {
+            description:
+                'Checks existence of local dependencies in the package.json',
+            category: 'Best Practices',
+            recommended: true
+        },
+        fixable: null // or "code" or "whitespace"
+    },
+
+    create: function(context) {
+        return {
+            'Program:exit': node => {
+                if (!isPackageJson(context.getFilename())) {
+                    return;
+                }
+                const sourceCode = context.getSourceCode();
+                const packageRoot = extractPackageObjectFromAST(node);
+                const original = JSON.parse(sourceCode.getText(packageRoot));
+                const {
+                    dependencies,
+                    peerDependencies,
+                    devDependencies
+                } = original;
+
+                const depObjs = [
+                    Object.entries(dependencies || {}),
+                    Object.entries(peerDependencies || {}),
+                    Object.entries(devDependencies || {})
+                ];
+                depObjs.map(obj =>
+                    obj.map(([key, value]) => {
+                        if (value.includes('link:')) {
+                            const filePath = path.join(
+                                context
+                                    .getFilename()
+                                    .replace(/package\.json/g, '/'),
+                                value.replace(/link:/g, ''),
+                                '/package.json'
+                            );
+
+                            try {
+                                if (!fileExistsWithCaseSync(filePath)) {
+                                    context.report({
+                                        node: packageRoot,
+                                        message: `The package ${key} does not exist given the specified path: ${value}.`
+                                    });
+                                }
+                            } catch (e) {
+                                context.report({
+                                    node: packageRoot,
+                                    message: `The package ${key} does not exist given the specified path: ${value}.`
+                                });
+                            }
+                        }
+                    })
+                );
+            }
+        };
+    }
+};
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "test": "nyc --reporter=lcov --reporter=text mocha tests --recursive",
     "test:debug": "mocha --inspect-brk tests --recursive",
     "test:watch": "mocha tests --recursive --watch",
+    "lint": "yarn run eslint --ext .json,.js . ",
     "pretest": "npm run format",
     "format": "prettier --ignore-path .gitignore --write \"**/*.{js,css,md}\" && eslint ./**/*.js"
   },
--- a/tests/lib/__fixtures__/local-dependency/package.json
+++ b/tests/lib/__fixtures__/local-dependency/package.json
@@ -1,0 +1,53 @@
+{
+    "name": "eslint-plugin-package-json",
+    "version": "0.1.3",
+    "description": "Rules for valid, consistent, and readable package.json files",
+    "keywords": [
+      "eslint",
+      "eslintplugin",
+      "eslint-plugin",
+      "magento"
+    ],
+    "author": "Magento Commerce",
+    "license": "(OSL-3.0 OR AFL-3.0)",
+    "main": "lib/index.js",
+    "homepage": "https://github.com/zetlen/eslint-plugin-package-json#readme",
+    "repository": {
+      "type": "git",
+      "url": "git+https://github.com/zetlen/eslint-plugin-package-json/issues"
+    },
+    "bugs": {
+      "url": "https://github.com/zetlen/eslint-plugin-package-json/issues"
+    },
+    "scripts": {
+      "test": "nyc --reporter=lcov --reporter=text mocha 'tests/lib/rules/local-dependency.js' --recursive",
+      "test:debug": "mocha --inspect-brk tests --recursive",
+      "test:watch": "mocha tests --recursive --watch",
+      "pretest": "npm run format",
+      "format": "prettier --ignore-path .gitignore --write \"**/*.{js,css,md}\" && eslint ./**/*.js"
+    },
+    "devDependencies": {
+      "eslint": "^5.8.0",
+      "lodash": "^4.17.15",
+      "mocha": "^6.2.0",
+      "nyc": "^14.1.1",
+      "prettier": "^1.18.2"
+    },
+    "dependencies": {
+      "disparity": "^3.0.0",
+      "package-json-validator": "^0.6.3",
+      "requireindex": "^1.2.0"
+    },
+    "peerDependencies": {
+      "eslint": ">=4.7.0"
+    },
+    "engines": {
+      "node": ">=8.0.0"
+    },
+    "nyc": {
+      "include": [
+        "lib/**"
+      ]
+    }
+  }
+  
--- a/tests/lib/rules/local-dependency.js
+++ b/tests/lib/rules/local-dependency.js
@@ -1,0 +1,299 @@
+/**
+ * @fileoverview Checks existence of local dependencies in the package.json
+ * @author Kendall Gassner
+ */
+'use strict';
+
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+const path = require('path');
+var rule = require('../../../lib/rules/local-dependency'),
+    RuleTester = require('eslint').RuleTester;
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const fileName = partialPath => {
+    return path.join(process.cwd(), partialPath);
+};
+
+var ruleTester = new RuleTester();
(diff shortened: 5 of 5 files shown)


lib/rules/local-dependency.js
```diff
@@ -0,0 +1,91 @@
...
+                } = original;
+
+                const depObjs = [
+                    Object.entries(dependencies || {}),
+                    Object.entries(peerDependencies || {}),
+                    Object.entries(devDependencies || {})
+                ];
+                depObjs.map(obj =>
+                    obj.map(([key, value]) => {
+                        if (value.includes('link:')) {
```
adierkens, line 63: Would it be worth adding in `file://` support to this as well?

kendallgassner:
Hi @zetlen! Happy new year!  Any chance you can take a look at this.

README.md
```diff
@@ -52,5 +52,20 @@ Or, individually configure the rules you want to use under the rules section.
...
+    "devDependencies": {
+            "some-package": "link:../folder",
+    }
+    ```
+    Should be:
+    ```
+    "devDependencies": {
+            "some-package": "link:../Folder",
+    }
+    ```
```
JoshuaKGoldberg, line 68: [Docs]

* It's much more common for npm projects to have lowercased package names. What do you think about switching the order here so that `link:../folder` is mentioned as preferred?
* This explanation confused me at first because it never explains _why_ one should be preferred. Can you add some mention of what the folder case is correctly?
    * If there's a good docs link, that might be nice too?

README.md
```diff
@@ -52,5 +52,20 @@ Or, individually configure the rules you want to use under the rules section.
 -   [`package-json/order-properties`](docs/rules/order-properties.md): Require top-level properties to be in a conventional order (`"name"`first, etc.).
 -   [`package-json/sort-collections`](docs/rules/sort-collections.md): Keep sub-collections like `"dependencies"` and `"scripts"` in alphabetical order.
 -   [`package-json/valid-package-def`](docs/rules/valid-package-def): Validate `package.json` files against the NPM specification.
+-   [`package-json/valid-package-def`](docs/rules/valid-package-def): Validates the casing for `file:` and `link:` dependencies in the `package.json` files. If the casing of these paths are incorrect yarn does not sub out the link when releasing. An example:
```
JoshuaKGoldberg, line 55: [Typo]

```suggestion
-   [`package-json/local-dependency`](docs/rules/local-dependency): Validates the casing for `file:` and `link:` dependencies in the `package.json` files. If the casing of these paths are incorrect yarn does not sub out the link when releasing. An example:
```

lib/rules/local-dependency.js
```diff
@@ -0,0 +1,99 @@
...
+const {
+    isPackageJson,
+    extractPackageObjectFromAST
+} = require('../processors/PackageJsonProcessor');
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+function fileExistsWithCaseSync(filepath) {
```
JoshuaKGoldberg, line 17: [Suggestion] Resolving a path to a package can be pretty tricky 😄. I would think `require.resolve` would work instead of the body of this function, to avoid a recursive call. Have you tried that?

Note that ESM environments might need https://nodejs.org/api/module.html#modulecreaterequirefilename...

lib/rules/local-dependency.js
```diff
@@ -0,0 +1,99 @@
...
+                    Object.entries(devDependencies || {})
+                ];
+
+                depObjs.map(obj =>
+                    obj.map(([key, value]) => {
+                        const response = localPath => {
+                            const filePath = path.join(
+                                context
+                                    .getFilename()
+                                    .replace(/package\.json/g, '/'),
```
JoshuaKGoldberg, line 68: [Testing] Could you add a test case for the directory path containing `package.json`, such as a path like `path/to/package.json/gotcha/package.json`? It's a rare edge case, but not impossible.

lib/rules/local-dependency.js
```diff
@@ -0,0 +1,99 @@
...
+                                });
+                            }
+                        };
+
+                        if (value.includes('link:')) {
+                            response('link:');
+                        }
+                        if (value.includes('file:')) {
+                            response('file:');
+                        }
```
JoshuaKGoldberg, line 93: [Nitpick] Very small, non-blocking: maybe `.startsWith` to be more precise?

tests/lib/__fixtures__/local-dependency/package.json
JoshuaKGoldberg, line 1: [Cleanup] This fixture file has a lot more metadata than the others. Is there a reason for most of it that I'm missing? If not, could you remove the unnecessary bits?

tests/lib/rules/local-dependency.js
```diff
@@ -0,0 +1,337 @@
...
+    valid: [
+        {
+            code: `module.exports = {
+                    "name": "pandages",
+                    "version": "1.0.0",
+                    "description": "",
+                    "main": "index.js",
+                    "keywords": [],
+                    "author": "me!",
+                    "license": "ISC",
```
JoshuaKGoldberg, line 33: [Cleanup] The other test files include a lot of unnecessary fields, but especially for this one with a lot of tests (which is awesome btw, love the test coverage!) I'm finding it hard to read through them. Would you be up for removing the unnecessary-for-tests fields?

```suggestion
```

package.json
```diff
@@ -23,6 +23,7 @@
     "test": "nyc --reporter=lcov --reporter=text mocha tests --recursive",
     "test:debug": "mocha --inspect-brk tests --recursive",
     "test:watch": "mocha tests --recursive --watch",
+    "lint": "yarn run eslint --ext .json,.js . ",
```
JoshuaKGoldberg, line 26: [Cleanup] Local testing artifact? 😄 

```suggestion
```

lib/rules/local-dependency.js
JoshuaKGoldberg, line 1: > `local-dependency`

[Naming] I'm not sold on this name 🤔. It's accurate, but not precise... Maybe `local-dependency-exists`? `valid-local-dependency`? I don't like either of those too much either, but would generally be interested in _something_ that conveys more precisely that this checks for the local dependency name existing. What do you think?

JoshuaKGoldberg (changes requested):
Cool, this is definitely useful! Requested a few changes here and there and left some questions/suggestions too. ✨

lib/rules/local-dependency.js
kendallgassner: I'll go with `valid-local-dependency` since it follows the naming convention of `valid-package-def`

README.md
```diff
@@ -52,6 +52,22 @@ Or, individually configure the rules you want to use under the rules section.
 -   [`package-json/order-properties`](docs/rules/order-properties.md): Require top-level properties to be in a conventional order (`"name"`first, etc.).
 -   [`package-json/sort-collections`](docs/rules/sort-collections.md): Keep sub-collections like `"dependencies"` and `"scripts"` in alphabetical order.
 -   [`package-json/valid-package-def`](docs/rules/valid-package-def.md): Validate `package.json` files against the NPM specification.
+-   [`package-json/local-dependency`](docs/rules/local-dependency.md): Validates the casing for `file:` and `link:` dependencies in the `package.json` files. If the casing of these paths are incorrect yarn does not sub out the link when releasing. An example:
```
kendallgassner, line 55:
```suggestion
-   [`package-json/valid-local-dependency`](docs/rules/valid-local-dependency.md): Validates the casing for `file:` and `link:` dependencies in the `package.json` files. If the casing of these paths are incorrect yarn does not sub out the link when releasing. An example:
```

lib/rules/valid-local-dependency.js
```diff
@@ -0,0 +1,85 @@
...
+                            const filePath = path.join(
+                                context
+                                    .getFilename()
+                                    .replace(/package\.json/g, '/'),
+                                value.replace(new RegExp(localPath, 'g'), ''),
+                                '/package.json'
+                            );
+
+                            try {
+                                if (!require.resolve(filePath)) {
```
kendallgassner, line 60: Thanks for the tip @JoshuaKGoldberg! I am not totally following the create resolve comment but it seems like require.resolve is working.

JoshuaKGoldberg: I was thinking that if & when ESLint supports running rules as ESM, this might crash. But I suppose we can cross that bridge when we get to it.

docs/rules/valid-local-dependency.md
```diff
@@ -0,0 +1,25 @@
+# Ensures that local dependencies specified in a package.json exist
+
+## Rule Details
+
+This rule validates the casing for `file:` and `link:` dependencies in a `package.json` file. If the casing of a `link:` or `file:` path is incorrect yarn does not sub out the link when releasing.
```
JoshuaKGoldberg, line 5: [Docs] It validates more than casing, no? Like, if the file doesn't exist at all, this will catch that, right?

Maybe:

```suggestion
This rule validates the path for `file:` and `link:` dependencies in a `package.json` file - including name casing. If a `link:` or `file:` path is incorrect yarn does not sub out the link when releasing.
```

docs/rules/valid-local-dependency.md
```diff
@@ -0,0 +1,25 @@
...
+
+Examples of **incorrect** code for this rule:
```
JoshuaKGoldberg, line 7: [Docs] I know this doesn't quite adhere to the normal rule docs convention, but it'd be good to mention what the right path is... maybe something like this?

```suggestion
Examples of **incorrect** code for this rule (when `../folder` is the correct path):
```

(and similar for the correct code)

lib/index.js
```diff
@@ -31,7 +31,8 @@ module.exports.configs = {
         rules: {
             'package-json/sort-collections': 'error',
             'package-json/order-properties': 'warn',
-            'package-json/valid-package-def': 'error'
+            'package-json/valid-package-def': 'error',
+            'package-json/local-dependency': 'error'
```
JoshuaKGoldberg, line 35:
```suggestion
            'package-json/valid-local-dependency': 'error'
```

package-lock.json
JoshuaKGoldberg, line 1: Heh, this keeps happening to me too. Hoping to get #33 / #39 working Soon™️. I'll try to remember to revert before merge.

For reference, `npx npm@6` instead of `npm` worked on my machine.

lib/rules/valid-local-dependency.js
```diff
@@ -0,0 +1,85 @@
...
+                } = original;
+
+                const depObjs = [
+                    Object.entries(dependencies || {}),
+                    Object.entries(peerDependencies || {}),
+                    Object.entries(devDependencies || {})
+                ];
+
+                depObjs.map(obj =>
+                    obj.map(([key, value]) => {
```
JoshuaKGoldberg, line 49: [Nitpicking] These would be a little more semantically correct as `forEach` (https://eslint.org/docs/latest/rules/array-callback-return). Definitely not a blocker IMO 😄 

```suggestion
                depObjs.forEach(obj =>
                    obj.forEach(([key, value]) => {
```

lib/rules/valid-local-dependency.js
```diff
@@ -0,0 +1,85 @@
...
+                                if (!require.resolve(filePath)) {
+                                    context.report({
+                                        node: packageRoot,
+                                        message: `The package ${key} does not exist given the specified path: ${value}.`
+                                    });
+                                }
+                            } catch (e) {
+                                context.report({
+                                    node: packageRoot,
+                                    message: `The package ${key} does not exist given the specified path: ${value}.`
```
JoshuaKGoldberg, line 69: [Thoughts] It'd be nice if it used a different, more explicit message if the only difference is casing. But that's probably better as a follow-up issue. I bet it's surprisingly hard to get right 😄 .

JoshuaKGoldberg (approved):
Awesome, thanks @kendallgassner! I'm not used to coming across three year old PRs and the original author immediately responding to review 😄 this was wonderful. 

Will leave open for a bit in case @zetlen wants to give input. Setting myself a reminder to check back in next week.

docs/rules/valid-local-dependency.md
JoshuaKGoldberg, line 5: > yarn

[Thoughts] I wonder what pnpm does here? And bun? Deno? I haven't looked into them and don't have plans to. 😄

lib/rules/valid-local-dependency.js
zetlen: At some point Node will support https://nodejs.org/api/esm.html#importmetaresolvespecifier so you could do a forward-compatible feature detection here.

zetlen:
@JoshuaKGoldberg You have my general signoff going forward with this and the rest of the code! Don't hesitate to ask me questions, but don't worry about my rubber stamp.

docs/rules/valid-local-dependency.md
kendallgassner: 🤔 good point -- I could remove the yarn explanation all together to generalize the rule. I originally wrote this because I kept messing up casing in my `yarn` project.

lib/rules/valid-local-dependency.js
kendallgassner: _Technically_ it will catch spelling mistakes and non-existing packages.... but package managers should catches those so the rule is mostly useful for casing errors.

JoshuaKGoldberg: Good point. I think I'd personally still interpret it for both, since it gives the nice errors before package managers would.

JoshuaKGoldberg:
Just a heads up @kendallgassner, I changed the license for future versions of the library to MIT in #38. That doesn't impact reviewing this PR at all - I just don't know how changing a license does/should impact existing PRs.

PR #12, opened 2021-01-03
fix: Missing extension in valid-package-def link

nschonni:


JoshuaKGoldberg:
Oh, shoot, sorry - missed this before I sent #27. But yes looks good, thanks!

PR #21, opened 2022-05-03
Fix broken link

aconrad:
Clicking that link was causing a 404.


JoshuaKGoldberg:
Dup of #12 - thanks though!

PR #30, opened 2023-09-13
fix: detect package.json in Windows

### LINKED ISSUE #13: Does not detect package.json in Windows environment
The culprit is the following line:

```
const isPackageJson = filePath =>
    filePath.endsWith('/package.json') || filePath === 'package.json';
```

filePath in my case is `C:\Users\bla\workspace\project\package.json` and does not match due to the `/` in your `endsWith()` test.

JoshuaKGoldberg:
Fixes #13

### DIFF
--- a/lib/processors/PackageJsonProcessor.js
+++ b/lib/processors/PackageJsonProcessor.js
@@ -1,7 +1,6 @@
 const parseErrors = {};
 
-const isPackageJson = filePath =>
-    filePath.endsWith('/package.json') || filePath === 'package.json';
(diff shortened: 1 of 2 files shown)


Zamiell:
wouldn't it be simpler to use [`path.sep`](https://nodejs.org/api/path.html#pathsep)?

JoshuaKGoldberg:
Probably. But the regexp avoids a dependency on `path`. Which might long-term be more portable.

Zamiell:
historically I've not thought of node builtins as "dependencies", but fair enough =p

PR #40, opened 2023-09-13
feat!: ESLint 8 and jsonc-eslint-parser

### LINKED ISSUE #39: Bump ESLint dependency to 8.x
Splitting out of #33: per https://eslint.org/docs/latest/extend/custom-processors#file-extension-named-processor, this package's file extension based processor approach is now deprecated. Fun!

### LINKED ISSUE #43: Tooling: consider switching to jsonc-eslint-parser
@ota-meshi's https://github.com/ota-meshi/jsonc-eslint-parser looks _excellent_: well-documented, well-tested, and well-used by https://www.npmjs.com/package/eslint-plugin-jsonc.

I imagine this might unblock #40? At the very least, I like the idea of using the standard community method for working with JSON.

JoshuaKGoldberg:
Fixes #39; fixes #43.

Moves to the external `jsonc-eslint-parser` because:

* #43: it's a good ecosystem standard
* ESLint's RuleTester doesn't support this usage of processors (https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/40#issuecomment-1726969178)

Also removes the `disparity` dependency, in favor of directly telling users to run the auto-fixer. This is the strategy other plugins such as [`eslint-plugin-simple-import-sort`](https://github.com/lydell/eslint-plugin-simple-import-sort) take. It's simpler to implement, makes for easier-to-read error messages, and reduces the size of `node_modules`.


JoshuaKGoldberg:
From the ESLint Discord: https://discord.com/channels/688543509199716507/711686958635548723/1151961528011063296

> Question: I just started overhauling an old plugin that uses a processor. I'd like to get it to use RuleTester for the rules. Is that blocked on https://github.com/eslint/rfcs/pull/31 (feat: RuleTester supports processor)?

> Response: Short answer: yes, you're blocked. Longer answer: it's not clear that the RFC is the right solution to this problem, and because you're not using JS, you are in a sort of "voiding your warranty" situation. 🙂
> It's possible that language plugins would make this easier to solve, but there's no straightforward answer right now.

Welp.

README.md
```diff
@@ -4,47 +4,46 @@ Rules for valid, consistent, and readable package.json files
...
-```
-
-Use the prepackaged config by adding an "extends" property, or appending to an existing "extends" property:
-
-```json
-{
-    "extends": ["eslint:recommended", "plugin:package-json/recommended"],
-    "plugins": ["package-json"]
-}
+Add an override to your ESLint configuration file that specifies this plugin, [`jsonc-eslint-parser`](https://github.com/ota-meshi/jsonc-eslint-parser) and its recommended rules for your `package.json` file:
```
JoshuaKGoldberg, line 17: fyi @ota-meshi - I ended up rolling #43 into this PR too.

Issue #31, opened 2023-09-13
order-properties: consider an option for sort-package-json's order


JoshuaKGoldberg:
Right now, the `package-json/order-properties` rule uses its own "standard order": https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/b6553eca1e314d9def74e37af94913785804d26e/lib/rules/order-properties.js#L16

I haven't seen this order in recent history. But, I _have_ seen https://github.com/keithamus/sort-package-json be used in recent ecosystem projects such as https://github.com/matzkoh/prettier-plugin-packagejson. Perhaps `order-properties` could take in an option?

cc @zetlen - is there a good reference for the current order? I'm wondering what to name if it there is an option added...

cc @kellyselden from https://github.com/kellyselden/eslint-plugin-json-files/blob/3a24e6f36b7828210a56070af4487cb5831f7f24/docs/rules/sort-package-json.md - I do find it interesting that both of these ESLint plugins have package.json sorting implemented...

Zamiell:
I think that most of the ecosystem uses `sort-package-json`.
I would vote that we match the ordering of that by default.

zetlen:
I think I took the ordering from the order in which [the documentation page describes them.](https://docs.npmjs.com/cli/v10/configuring-npm/package-json) It doesn't say anywhere that this is a _required_ order. If there's another, more widespread convention, use that.

JoshuaKGoldberg:
Awesome! Let's go with, say, `order: "legacy" | "sort-package-json"``? With `"legacy"` as the default now, noted to switch to `"sort-package-json"` in a potential future version?

Marking as accepting PRs. If nobody's sent a PR by the time I've overhauled this repo (#33 through #37) I'll go ahead and send one myself.

Note that adding `sort-package-json` as a dependency might clobber `package-lock.json` until #33 is in. I've found that using `npx npm@6` to run an older version of npm is a workaround.

kendallgassner:
@JoshuaKGoldberg can you assign this ticket to me. I am going to take a look today.

Issue #33, opened 2023-09-13
Bump dev dependencies to latest


JoshuaKGoldberg:
Splitting out of #28: it's been a while since the `package.json` dependencies were updated. I'd like to get them all bumped to the latest.

JoshuaKGoldberg:
Oh boy, file extension based processors are deprecated in ESLint now. https://eslint.org/docs/latest/extend/custom-processors#file-extension-named-processor -> #39.

Issue #37, opened 2023-09-13
Onboard to create-typescript-app


JoshuaKGoldberg:
I've been migrating a bunch of repos under my purview to https://github.com/JoshuaKGoldberg/create-typescript-app. It contains a bunch of IMO useful tooling and helps standardize on using "latest and greatest" configurations for them. Plus having automated all-contributors recognition, release, etc. is quite nice.

Confirmed with @zetlen that we're clear to onboard. But since CTA is quite new I'll definitely put forward the PR as a proposal first!

Will tackle after: #34, #35, #36.

JoshuaKGoldberg:
Marking as blocked on #35, which is blocked on #39.

PR #55, opened 2023-10-02
Update order-properties to use sort-package-json

kendallgassner:
closes: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/31

### What

- Replaced the `order-properties` sorting algorithm to use [sort-package-json](https://github.com/keithamus/sort-package-json).
- Changed the type of `options` from an `Array<string>` to 
```tsx
{
    order?: "legacy" | "sort-package-json" | Array<string>
}
```
- defaulted the sort order to `legacy`
- Add additional testing to test options: "legacy" | "sort-package-json" .

#### Option: `order: Array<string>`

Previously, any property that was not provided in the `order: Array<string>` collection was ignored by the sort algorithm. Now that we are using `sort-package-json`, [a property that is not provided in the `order: Array<string>` will be sorted by the `defaultSortOrder` provided by the library.](https://github.com/keithamus/sort-package-json#optionssortorder)

#### Sort-package-json version
 the latest version of sort-package-json is [an esm module](https://github.com/keithamus/sort-package-json/blob/main/package.json#L20) so I opted to use an earlier version to avoid heavy refactorign. This package can be bumped when #33 is merged.

### DIFF
--- a/docs/rules/order-properties.md
+++ b/docs/rules/order-properties.md
@@ -31,19 +31,28 @@ Examples of **correct** code for this rule:
 
 ### Options
 
-Pass an array of top-level package properties to lint sorting on only those
-properties. All properties not in this collection will be ignored.
-
-Example:
-
 ```js
-"package-json/order-properties": ["error", [
-    "name",
-    "version" // Ensure only that name precedes version
-]]
+"package-json/order-properties": ["error", {
+    order: "sort-package-json"
+}]
 ```
 
-Defaults:
+#### Order
+
+The `order` property specifies the sorting order of package properties. Pass in:
+
+-   "legacy" - to order properties specified by [npm documentation](https://docs.npmjs.com/cli/v10/configuring-npm/package-json).
+-   "sort-package-json" - to order properties by the default order specified in [sort-package-json](https://github.com/keithamus/sort-package-json).
+-   Array<string> - to specify an array of top-level package properties to lint sorting on only those
+    properties. All properties not in this collection will be sorted by "sort-package-json" specifications.
+
+```tsx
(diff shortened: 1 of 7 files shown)


docs/rules/order-properties.md
```diff
@@ -31,19 +31,28 @@ Examples of **correct** code for this rule:
...
-```js
-"package-json/order-properties": ["error", [
-    "name",
-    "version" // Ensure only that name precedes version
-]]
+The `order` property specifies the sorting order of package properties. Pass in:
+
+-   "legacy" - to order properties specified by [npm documentation](https://docs.npmjs.com/cli/v10/configuring-npm/package-json).
+-   "sort-package-json" - to order properties by the default order specified in [sort-package-json](https://github.com/keithamus/sort-package-json).
+-   Array<string> - to specify an array of top-level package properties to lint sorting on only those
```
JoshuaKGoldberg, line 46: [Docs] The last item shows up as _"Array - to specify ..."_ in GitHub's rendered pages. Best to backtick.

```suggestion
-   `"legacy"` - to order properties specified by [npm documentation](https://docs.npmjs.com/cli/v10/configuring-npm/package-json).
-   `"sort-package-json"` - to order properties by the default order specified in [sort-package-json](https://github.com/keithamus/sort-package-json).
-   `Array<string>` - to specify an array of top-level package properties to lint sorting on only those
```

lib/rules/sort-collections.js
JoshuaKGoldberg, line 1: Changes here look unrelated? I plan on normalizing formatting, etc in #37.

PR #64, opened 2023-11-06
chore: switch valid-local-dependency tests to not rely on file system casing

### LINKED ISSUE #60: Test failures on main with valid-local-dependency.js
If you run `npm test` on the `main` branch, everything passes. ✅ 

But if you specifically run the `valid-local-dependency` rule tests, they fail:

```shell
npx mocha tests/lib/rules/valid-local-dependency.js
```

```plaintext
  10 passing (37ms)
  8 failing
```

Specifically all but the fourth `invalid` cases are failing.

```plaintext
1) valid-local-dependency
       invalid
         {
                        "license": "ISC",
                        "dependencies": {
                            "some-package": "link:./tests/lib/__fixtures__/Valid-local-dependency",
                            "some-other-package": "some-other-package"
                        }
            }:

      AssertionError [ERR_ASSERTION]: Should have 1 error but had 0: []
      + expected - actual

      -0
      +1   
```

FYI @kendallgassner - I had no idea this was a "quirk" when reviewing #11. It's popping up in #35 by virtue of switching to Vitest and all tests now running in CI.

JoshuaKGoldberg:
Fixes #60.

The problem was the `invalid` cases weren't giving errors when the resolved path differed from file system casing. Requested paths like `.../Valid-local-dependency` were passing when the file case on system was `.../valid-local-dependency`. I played a bit with adding explicit logic to check for case sensitivity but:

* That can get complicated
* Not all operating systems are case sensitive by default

...so this instead switches the tests to use paths that differ in more than just casing.

### DIFF
--- a/tests/lib/rules/valid-local-dependency.js
+++ b/tests/lib/rules/valid-local-dependency.js
@@ -71,7 +71,7 @@ ruleTester.run('valid-local-dependency', rule, {
         {
             code: `{
                     "dependencies": {
-                        "some-package": "link:./tests/lib/__fixtures__/Valid-local-dependency",
+                        "some-package": "link:./tests/lib/__fixtures__/Invalid-local-dependency",
                     }
             }`,
             filename: fileName('not-package.json')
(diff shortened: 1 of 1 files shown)


kendallgassner:
Thank you for doing this!

kendallgassner (approved):

Issue #91, opened 2024-01-05
🚀 Feature: Support flat config


KristjanESPERANTO:
### Bug Report Checklist

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have pulled the latest `main` branch of the repository.
- [X] I have [searched for related issues](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues?q=is%3Aissue) and found none that matched my issue.

### Overview

Configuration using eslintrc [will be deprecated soon](https://eslint.org/blog/2023/10/flat-config-rollout-plans/#flat-config-by-default-in-eslint-v9.0.0). Unfortunately I can't manage to get this plugin into a flat configuration. Is this already possible?

JoshuaKGoldberg:
@all-contributors please add @kristjanesperanto for ideas.

> 🤖 Beep boop! This comment was added automatically by [all-contributors-auto-action](https://github.com/marketplace/actions/all-contributors-auto-action).
> Not all contributions can be detected from Git & GitHub alone. Please comment any missing contribution types this bot missed.
> ...and of course, thank you for contributing! 💙

JoshuaKGoldberg:
ESLint's [Plugin Migration Guide](https://eslint.org/docs/latest/extend/plugin-migration-flat-config) has more info. For this plugin in particular we'd want to:

* Expose the `meta` and `configs` properties
* Document how to use them with flat config in the README.md

KristjanESPERANTO:
Great! Thank you very much! :fire:

PR #157, opened 2024-02-19
refactor: move to type module

### LINKED ISSUE #152: 🐛 Bug: Can not import this in ESM package
### Bug Report Checklist

### Expected

No error with flat config

### Actual

```
TypeError: Cannot read properties of undefined (reading 'docs')
```

### Additional Info

Related: #91 

If you use this plugin like this:

```js
import * as pluginPackageJson from 'eslint-plugin-package-json'

{
  files: ['package.json'],
  plugins: {
    'package-json': pluginPackageJson,
  },
  rules: pluginPackageJson.configs.recommended.rules,
},
```

You will get an error, because flat config use esm. But in `lib/index.mjs`, it always use default import from a cjs file.

https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/baf1aaa2696c3da4634f21f0d091669b3d3fb97d/src/index.ts#L1-L9

The imported `validVersion` looks like

```
{
  "default": {
    "meta": {
      "docs": {
        "category": "Best Practices",
        "description": "Enforce that package versions are valid semver specifiers",
        "recommended": true
      },
      "hasSuggestions": true,
      "messages": {
        "invalid": "Version is not a valid semver specifier.",
        "type": "\"version\" should be a string."
      }
    }
  }
}
```

So `rule.meta` is `undefined`

https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/baf1aaa2696c3da4634f21f0d091669b3d3fb97d/src/index.ts#L26-L28

hyoban:
## PR Checklist

-   [x] Addresses an existing open issue: part of #91 

## Overview

This refactor allows us to use flat config now, although no default preset is provided yet. fix #152 

```js
import * as pluginPackageJson from 'eslint-plugin-package-json'
import parserJsonc from 'jsonc-eslint-parser'

export default [
  {
    files: ['package.json'],
    languageOptions: {
      parser: parserJsonc,
    },
    plugins: {
      'package-json': pluginPackageJson,
    },
    rules: pluginPackageJson.configs.recommended.rules,
  },
]
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,18 +12,18 @@
 		"name": "James Zetlan",
 		"email": "zetlan@gmail.com"
 	},
-	"type": "commonjs",
+	"type": "module",
 	"exports": {
 		".": {
 			"types": {
-				"import": "./lib/index.d.mts",
-				"require": "./lib/index.d.ts"
+				"import": "./lib/index.d.ts",
+				"require": "./lib/index.d.cts"
 			},
-			"import": "./lib/index.mjs",
-			"require": "./lib/index.js"
+			"import": "./lib/index.js",
+			"require": "./lib/index.cjs"
 		}
 	},
-	"main": "./lib/index.js",
+	"main": "./lib/index.cjs",
 	"files": [
 		"lib/",
 		"package.json",


package.json
JoshuaKGoldberg, line 1: 🤔 I'm confused. How does switching this to a module change flat config support?

I realize #91 doesn't have much info, so I just posted some there: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/91#issuecomment-1953437450.

JoshuaKGoldberg (changes requested):
Thanks for getting started on this! 🙌

package.json
hyoban: This is related to the extension in import.

Currently, why we can not use this plugin in flat config(esm) is explained in #152. After moving to `type: module`, we can correctly import the esm build.

JoshuaKGoldberg: Got it, so the issue wasn't a duplicate after all. Will re-open. Thanks!

hyoban:
I feel like using named imports is more of a compromise. We can now use `type : module` without dropping cjs support.

JoshuaKGoldberg:
Yeah I see what you're saying - it makes sense that named exports is a bit of a compromise. And you're right that we could switch to `type: module` and consumers would still be able to work with it. But I worry about it 😅 and would rather do that as a breaking change down the line.

Maybe: let's go with `named exports` for now, and file switching to type module for a future version?

hyoban:
Close this one for now

PR #161, opened 2024-02-20
feat: support flat config

### LINKED ISSUE #91: 🚀 Feature: Support flat config

hyoban:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #91

## Overview

Some code are not very good, looking forward to your suggestions

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -47,7 +47,7 @@ module.exports = {
 			},
 		},
 		{
-			files: "**/*.md/*.ts",
+			files: ["**/*.md/*.ts", "**/*.md/*.js"],
 			rules: {
 				"n/no-missing-import": [
 					"error",
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -47,7 +47,7 @@ module.exports = {
 			},
 		},
 		{
-			files: ["**/*.md/*.ts", "**/*.md/*.js"],
+			files: "**/*.md/*.ts",
 			rules: {
 				"n/no-missing-import": [
 					"error",
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ export default [
 
 If you want to override the recommended rules:
 
-```js
+```ts
 import packageJson from "eslint-plugin-package-json/configs/recommended";
 
 export default [
@@ -100,7 +100,7 @@ module.exports = {
 ### Usage Alongside Prettier
 
 **[`prettier-plugin-packagejson`](https://github.com/matzkoh/prettier-plugin-packagejson)** is a [Prettier plugin](https://prettier.io/docs/en/plugins) that enforces the same `package.json` keys ordering as the [`order-properties`](docs/rules/order-properties.md) and [sort-collections](docs/rules/sort-collections.md) rules with default options.
-We recommend using both the Prettier plugin and `extends: ["plugin:package-json/recommended"]`.
+We recommend using both the Prettier plugin and `eslint-plugin-package-json`'s recommended configuration.
 The default settings don't conflict, and Prettier plugins can quickly fix up ordering in your editor on save and/or as a Git hook.
 
 ## Supported Rules
--- a/README.md
+++ b/README.md
@@ -29,9 +29,11 @@ npm install eslint eslint-plugin-package-json jsonc-eslint-parser --save-dev
 
 ### Flat Config
 
+This plugin's recommended configuration enables its rules on `**/package.json` files, parsing them with [`jsonc-eslint-parser`](https://github.com/ota-meshi/jsonc-eslint-parser):
+
 In your ESLint configuration file:
 
-```js
+```ts
 import packageJson from "eslint-plugin-package-json/configs/recommended";
 
 export default [
@@ -56,6 +58,8 @@ export default [
 	},
 ];
 ```
+
+See [ESLint's _Configuration Files_ guide](https://eslint.org/docs/latest/use/configure/configuration-files-new) for details on how to customize your rules and other config settings.
 
 ### Legacy Config
 
--- a/README.md
+++ b/README.md
@@ -27,6 +27,38 @@ npm install eslint eslint-plugin-package-json jsonc-eslint-parser --save-dev
 
 ## Usage
 
+### Flat Config
+
+In your ESLint configuration file:
+
+```js
+import packageJson from "eslint-plugin-package-json/configs/recommended";
+
+export default [
+	// your other ESLint configurations
+	packageJson,
+];
+```
+
+If you want to override the recommended rules:
+
+```js
+import packageJson from "eslint-plugin-package-json/configs/recommended";
+
+export default [
+	// your other ESLint configurations
+	{
+		...packageJson,
+		rules: {
+			...packageJson.rules,
+			"package-json/valid-package-def": "off",
+		},
+	},
+];
+```
+
+### Legacy Config
+
 Add an override to your ESLint configuration file that specifies this plugin, [`jsonc-eslint-parser`](https://github.com/ota-meshi/jsonc-eslint-parser), and its recommended rules for your `package.json` file:
 
 ```js
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ The default settings don't conflict, and Prettier plugins can quickly fix up ord
 💼 Configurations enabled in.\
 ✅ Set in the `recommended` configuration.\
 🔧 Automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/user-guide/command-line-interface#--fix).\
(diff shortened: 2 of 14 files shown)


src/index.ts
```diff
@@ -8,6 +11,17 @@ import { rule as validPackageDef } from "./rules/valid-package-def.js";
 import { rule as validRepositoryDirectory } from "./rules/valid-repository-directory.js";
 import { rule as validVersion } from "./rules/valid-version.js";
 
+declare const TSUP_FORMAT: "cjs" | "esm";
+const req =
+	typeof TSUP_FORMAT === "undefined" || TSUP_FORMAT === "cjs"
+		? require
+		: createRequire(import.meta.url);
```
JoshuaKGoldberg, line 18: [Refactor] I'm not 100% confident in CJS/ESM shenanigans, but - my instinct is that encoding tsup-specific globals into the source code isn't super long-term tenable. I'd rather rely on checking what exists at build/run-time rather than details of the tool used to build. Is there any reason why the following check of _"either ESM's `import.meta.url` works or CJS's `__filename` will exist"_ wouldn't work?

```suggestion
const require = createRequire(import.meta.url || __filename);
```

src/index.ts
```diff
@@ -20,12 +34,53 @@ export const rules = {
...
 export const configs = {
 	recommended: {
-		rules: Object.fromEntries(
-			Object.entries(rules)
-				.filter(([, rule]) => rule.meta.docs?.recommended)
-				.map(([name]) => ["package-json/" + name, "error" as const]),
-		),
+		rules: recommendedRuleSettings,
+	},
+	["recommended-flat"]: {
```
JoshuaKGoldberg, line 72: "Flat" isn't a great long-term way to refer to the configs. ESLint will be deprecating eslintrc configs soon, so all configs will be flat. "Flat" as a term will be redundant.

My preference would be to expose them both as `configs.recommended`. As in, only having one config property. The new version is essentially a superset of the old one (both have the same `rules`), and when I tried unifying them locally it all seemed nice and working. https://github.com/solidjs-community/eslint-plugin-solid takes this approach too and it seems to be working.

If that doesn't work _(and I'd really want to know - why not?)_ then I'd prefer calling the old config `recommended-legacy` or something like that the way [`eslint-plugin-markdown`](https://github.com/eslint/eslint-plugin-markdown/pull/232) does.

src/index.ts
```diff
@@ -20,12 +34,53 @@ export const rules = {
...
+		files: ["**/package.json"],
+		languageOptions: {
+			parser: parserJsonc,
+		},
+		plugins: {
+			"package-json": plugin,
```
JoshuaKGoldberg, line 78: [Refactor] Heh, I think I see why you had to alternate between declaring `configs` and `plugins`. It's that there's a circular reference between `configs.plugins["recommended-flat"]["package-json"]` and `plugins.configs` right?

We can get around that with a getter:

 ```suggestion
				get "package-json"() {
				  return plugin;
				}
```

No more need for an `interface`, `as` assertion, or `Object.assign`. 😎

File: src/index.ts [hunk 4]
```diff
@@ -1,3 +1,6 @@
+import parserJsonc from "jsonc-eslint-parser";
```
JoshuaKGoldberg, line 1: [Bug] When I try this in a CommonJS config file (`eslint package.json --config eslint.config.cjs`), the default-imported module is `undefined`. `jsonc-eslint-parser` doesn't actually provide a `default` export.

```suggestion
import * as parserJsonc from "jsonc-eslint-parser";
```

src/index.ts
```diff
@@ -20,12 +34,53 @@ export const rules = {
 	"valid-version": validVersion,
 };
 
+const recommendedRuleSettings = Object.fromEntries(
+	Object.entries(rules)
+		.filter(([, rule]) => rule.meta.docs?.recommended)
+		.map(([name]) => ["package-json/" + name, "error" as const]),
+);
```
JoshuaKGoldberg, line 41: [Praise] I like this refactor 😄 nice.

JoshuaKGoldberg (changes requested):
Ok! So sorry for taking multiple weeks to get back to you on this - it's a good change and important for the plugin. Thank you very much for getting it started!

The general path you took here seems reasonable to me. Glad to see the actual changes to `src/index.ts` are pretty straightforward. I left some requests inline about streamlining how the dual-emit (CJS/ESM) and dual-format (flat/legacy) support should look.

Also, per https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/91#issuecomment-1953437450, requesting:

> Document how to use them with flat config in the README.md

I think I'd made a note to get back to this PR once that was in and forgot to follow up 😅.

src/index.ts
hyoban: This idea is great, I have refactored the code to use the same strategy as eslint-plugin-solid. I used `recommended-flat` before because using `recommended-legacy` may be a breaking change.

hyoban:
Hi, I have changed the code to apply your suggestion. I intentionally left the documentation section until the implementation is refined based on your review, which has now been added.

Thanks a lot for your review. I have learned a lot.

.eslintrc.cjs
```diff
@@ -47,7 +47,7 @@ module.exports = {
 			},
 		},
 		{
-			files: "**/*.md/*.ts",
+			files: ["**/*.md/*.ts", "**/*.md/*.js"],
```
JoshuaKGoldberg, line 50: [Refactor] Alternately, if the snippets are switched to `\`\`\`ts\`, there'll be no need for this.

```suggestion
			files: "**/*.md/*.ts",
```

README.md
```diff
@@ -27,6 +27,38 @@ npm install eslint eslint-plugin-package-json jsonc-eslint-parser --save-dev
 
 ## Usage
 
+### Flat Config
+
+In your ESLint configuration file:
+
+```js
```
JoshuaKGoldberg, line 34: [Style] Here and the other added block - per the comment in `.eslintrc.cjs`

```suggestion
```ts
```

src/plugin.ts
```diff
@@ -0,0 +1,44 @@
...
+import { rule as validVersion } from "./rules/valid-version.js";
+
+const require = createRequire(import.meta.url || __filename);
+
+const { name, version } = require("../package.json") as {
+	name: string;
+	version: string;
+};
+
+const allRules = {
```
JoshuaKGoldberg, line 20: [Naming] Super nitpicky, but: I found it confusing that this is named `allRules` and then elsewhere is always referred to as `rules`. What do you think about renaming to `rules`?

```suggestion
const rules = {
```

README.md
```diff
@@ -27,6 +27,38 @@ npm install eslint eslint-plugin-package-json jsonc-eslint-parser --save-dev
...
+	{
+		...packageJson,
+		rules: {
+			...packageJson.rules,
+			"package-json/valid-package-def": "off",
+		},
+	},
+];
+```
+
```
JoshuaKGoldberg, line 59: [Docs] Since the legacy config has a mention of ESLint docs, I feel like it'd be nice to mention one here too. Maybe:

```suggestion

See [ESLint's _Configuration Files_ guide](https://eslint.org/docs/latest/use/configure/configuration-files-new) for details on how to customize your rules and other config settings.

```

README.md
```diff
@@ -27,6 +27,38 @@ npm install eslint eslint-plugin-package-json jsonc-eslint-parser --save-dev
 
 ## Usage
 
+### Flat Config
+
```
JoshuaKGoldberg, line 31: [Docs] Since the new config does a lot more for users, I think we'll want to explain a bit of it to them. Maybe:

```suggestion

This plugin's recommended configuration enables its rules on `**/package.json` files, parsing them with [`jsonc-eslint-parser`](https://github.com/ota-meshi/jsonc-eslint-parser):

```

JoshuaKGoldberg (approved):
This looks great, thanks a bunch @hyoban! 🙌 

I only have some small, honestly non-blocking comments left. Do you have opinions on them?

If not, no worries, I've set a reminder to merge this on Wednesday. 🚀

hyoban:
@JoshuaKGoldberg I have no problem with your modification, but I don't have a computer with me now. You can modify and merge it directly. Or I will deal with it tomorrow.

JoshuaKGoldberg:
You got it, I'll go ahead and do that now so we can release today. Thanks!

src/configs/recommended.test.ts
JoshuaKGoldberg, line 1: Figured I'd add a quick one here so code coverage doesn't drop. 🤷

PR #315, opened 2024-04-29
fix: rule order-properties not work

hyoban:
## PR Checklist

## Overview

Just a quick fix. I don't know why this worked before.


src/rules/order-properties.ts
JoshuaKGoldberg, line 1: [Testing] Oh, nice catch - this only happens if there are multiple failures, right? Can you add a unit test please?

JoshuaKGoldberg (changes requested):
oopity

src/rules/order-properties.ts
hyoban: Added, just make sure the first key is corrected.

JoshuaKGoldberg (approved):
Swell, thanks!

PR #498, opened 2024-08-25
fix(valid-repository-directory): use repository root for more accurate linting

### LINKED ISSUE #252: 🐛 Bug: valid-repository-directory rule always fails
### Bug Report Checklist

### Expected

My repository.directory field is valid so I expect the lint to pass

### Actual

The valid-repository-directory rule _always_ errors for me

### Additional Info

I debugged the lint rule and here's what I found:

| Variable | Value |
|-|-|
| context.filename | '/Volumes/sourcecode/npm-packages/packages/html-cdnify/package.json' |
| expected | '/Volumes/sourcecode/npm-packages/packages/html-cdnify' |
| directoryValue | 'packages/html-cdnify' |
| path.normalize(directoryValue) | 'packages/html-cdnify' |

And so this line always fails, for all my packages:

```js
if (path.normalize(directoryValue) !== expected) {
```

I see from all the tests that you assume `context.filename` is a relative path. So I'm not sure if my setup is different or weird in some way (source code on a volume? version of eslint?) or if this assumption is just wrong.

If the assumption is wrong then just let me know and I'll make the fix and submit a PR, assuming that is welcome. I believe the fix is to only check that the end of the path matches the repository.directory value, which is a little loose, but the more rigorous approach is to determine the root of the git repository, check that it has a remote that matches the one in the repository field, and then to compare using the repository root, and I assume we can't reasonably do that in a lint rule:

```js
if (!expected.endsWith(path.normalize(directoryValue))) {
```

altano:
## PR Checklist

-   [x] Fixes #252

### Commits

There are two commits:

~1. "settings" - these are just ignores I'm adding to keep my local development environment out of this project (e.g. I use devbox to manage my tool chain, so I'm ignoring those files from the repo)~
2. "fix ..." - the actual lint rule fixes

### @altano/repository-tools Dependency

The code changes here aren't that big because we're leaning on my [@altano/repository-tools](https://www.npmjs.com/package/@altano/repository-tools) library to do the hardest part: discovering the root of the repository.

Notes about this package:
- SCMs supported: git / mercurial / sapling / subversion
- Performance: the library can find a git repository's root in ~3ms, and the worst (slowest) repository type (Mercurial) takes ~148ms. I think this is totally acceptable for package.json linting:

  |operation   |scm         |hz     |min    |max    |mean   |p75    |p99    |p995   |p999   |rme     |
  |------------|------------|-------|-------|-------|-------|-------|-------|-------|-------|--------|
  |findRootSync|git         |294.36 |3.0661 |5.4844 |3.3972 |3.4789 |4.3417 |5.4844 |5.4844 |±1.37%  |
  |findRootSync|mercurial   |6.7348 |146.39 |152.29 |148.48 |149.20 |152.29 |152.29 |152.29 |±0.91%  |
  |findRootSync|sapling     |28.1276|33.7934|37.6630|35.5523|36.0934|37.6630|37.6630|37.6630|±1.55%  |
  |findRootSync|subversion  |69.9382|13.3270|18.7447|14.2983|14.2685|18.7447|18.7447|18.7447|±2.08%  |
- This repository is well tested. As part of the CI flow it has unit tests that create real repositories for each supported type and fully exercise the code.

If you don't want to depend on this library, we can pull in the very small amount of code directly into this repo, but you'd be losing out on the really good CI testing.

### Rule Fix

When in a supported repository, the logic for the lint rule is simple:
- Get the repository root path
- We calculate the relative path from the repository root to the directory of the package.json file being linted
- The package.json `"directory"` field should be that path
- If it isn't, we suggest it
- Example:
  - repository root: "~/src/project"
  - package.json file being linted: "~/src/project/packages/packageA/package.json"
  - the expected value for `"directory"` in package.json is `"packages/packageA"`

When a repository root cannot be found, there is a fallback:
- We make sure the `"directory"` value appears at the end of the package.json directory path. Since we don't know the root of the repository, we can't tell how much of the path has to be present. This can be inaccurate. For example, if package.json is at '/a/b/c/package.json':
  - We make sure `"directory"` is either "/a/b/c" or "b/c" or "c" (these all validate)
  - If you specify a directory that doesn't appear in this path, e.g. 'd', validation fails
  - If you specify something from the middle of the path, e.g. just 'b', validation fails (we _could_ relax the rule and allow this to pass, but I can't think of a situation in which this would be valid)
- There are no suggestions in this fallback mode.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -9,6 +9,7 @@
 		"pnpm-lock.yaml"
 	],
 	"words": [
+		"altano",
 		"autofixable",
 		"Codecov",
 		"codespace",
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
 		"*": "prettier --ignore-unknown --write"
 	},
 	"dependencies": {
+		"@altano/repository-tools": "^0.1.1",
 		"detect-indent": "6.1.0",
 		"detect-newline": "3.1.0",
 		"package-json-validator": "^0.6.5",
--- a/src/rules/valid-repository-directory.ts
+++ b/src/rules/valid-repository-directory.ts
@@ -1,10 +1,38 @@
 import type { AST as JsonAST } from "jsonc-eslint-parser";
 
+import { findRootSync } from "@altano/repository-tools/findRootSync.cjs";
 import * as ESTree from "estree";
 import * as path from "node:path";
 
 import { createRule } from "../createRule.js";
 import { findPropertyWithKeyValue } from "../utils/findPropertyWithKeyValue.js";
+
+/**
+ * Checks if the child path appears at the end of the parent path. e.g.
+ *
+ *   '/a/b/c', 'c' => true
+ *   '/a/b/c', 'b/c' => true
+ *   '/a/b/c', 'b' => false
+ *   '/a/b/c', 'd' => false
+ */
+function pathEndsWith(parent: string, child: string): boolean {
+	const segments = parent.split(path.sep);
+
+	if (parent === child) {
+		// the directory specified was the full, absolute path to the
+		// package.json
+		return true;
+	}
+
+	// work backwards from the end, adding another path segment to each check
+	let pathToCheck = "";
+	return segments.reverse().some((segment) => {
+		pathToCheck = path.join(segment, pathToCheck);
+		if (pathToCheck === child) {
+			return true;
+		}
+	});
+}
 
 export const rule = createRule({
 	create(context) {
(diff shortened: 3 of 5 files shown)


altano:
@JoshuaKGoldberg I left this PR in "draft" mode because I wanted to ask if you knew how to test this as an actual ESLint rule? There are the unit tests but I would love to test this in a real repo if possible. Any clue? Or can we publish a beta or something to npm and I'll test it that way? Once it's better tested I can publish.

altano:
@JoshuaKGoldberg OH, shoot:

The tests are failing because I published @altano/repository-tools as an ESM-only package, and this lint rule has a CJS build which fails to run b/c of the error: `Error [ERR_REQUIRE_ESM]: require() of ES Module ...`

We have options to proceed here:
1. I re-publish @altano/repository-tools as a dual CJS / ESM module, which I would really, _really_ love to avoid doing.
2. Node.js added support for requiring ESM modules, but it's behind an experimental flag even in 22.7.0: https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require, so I assume that's a non-starter (for now)
3. This package could be published ESM-only
4. Something else?

Is 1 the only viable option? Thoughts?

altano:
Alright, yeah, I don't see any other way forward, so I'll publish a dual package.

w.r.t. package surface area: I was planning on adding more utility functions to the "repository-tools" package, but only ones that are thin wrappers around the CLI commands. For example, the `findRootSync` method will (in a git repo) just run `git rev-parse --quiet --show-toplevel` to get the result (with equivalents for other SCM types). So it's very light on business logic and will forever remain so. But I'm happy to publish separate, smaller packages and then an umbrella one, so it's totally your call and I'm happy to oblige. Just LMK!

altano:
@JoshuaKGoldberg I investigated the best way to break a monolithic package up. Packages like lodash seem to have gone through multiple methods and landed on the approach of having one package, but allowing subpath exports into the individual files. Then you get the best of all worlds: it's tree-shaking friendly, but shared code won't be duplicated across sub-packages.

So, in addition to supporting ESM/CJS, I'm adding sub-path exports so you can do this:

```ts
import { findRootSync } from "@altano/repository-tools/findRootSync";
```

I also added unit tests to actually exercise the sub-path exports and make sure they work.

Does this all sound good to you? To reiterate: I'm down for whatever, so feel free to suggest an alternative.

JoshuaKGoldberg:
Cool, that sounds good to me!

I'm not super picky right now. Even if we somehow get it horribly wrong now, as you said, the package is pretty small. We can always tweak it later. 🙂

src/rules/valid-repository-directory.ts
```diff
@@ -26,24 +54,77 @@ export const rule = createRule({
...
+						!pathEndsWith(
+							fileDirectory,
+							path.normalize(directoryValue),
+						)
+					) {
+						context.report({
+							messageId: "mismatched",
+							node: directoryProperty.value as unknown as ESTree.Node,
+							// we don't have a good suggestion to offer
+							suggest: [],
```
JoshuaKGoldberg, line 87: Nit: I think this is implied?

```suggestion
```

src/tests/rules/valid-repository-directory.test.ts
```diff
@@ -1,123 +1,180 @@
...
+		},
+		{
+			code: `{ "repository": { "directory": "/Users/face/src/faceee" } }`,
+			filename: `/Users/face/src/faceee/package.json`,
+			name: `full, absolute path to package.json`,
+		},
+	],
+});
+
+const thisRepoDirectory = path.resolve(__dirname, "..", "..", "..");
```
JoshuaKGoldberg, line 105: Nit: any reason not to have a single string? 

```suggestion
const thisRepoDirectory = path.resolve(__dirname, "../../..");
```

package.json
```diff
@@ -58,6 +58,7 @@
 		"*": "prettier --ignore-unknown --write"
 	},
 	"dependencies": {
+		"@altano/repository-tools": "^0.1.1",
```
JoshuaKGoldberg, line 61: > Performance: the library can find a git repository's root in ~3ms, and the worst (slowest) repository type (Mercurial) takes ~148ms. I think this is totally acceptable for package.json linting:

Agreed. I have yet to meet someone who uses Mercurial day-to-day and is thrilled enough about it to want to optimize that seventh of a second.

JoshuaKGoldberg (approved):
OK! This is phenomenal, awesome job @altano! I tried this out in a few repositories locally. I couldn't break it at all. Nice! 👏 

I just have little nitpicky questions that don't really change much. I can go ahead and apply them in a couple days.

src/tests/rules/valid-repository-directory.test.ts
altano: hard-coding `path.sep` as `/` could lead to bugs on Windows

.eslintignore
```diff
@@ -1,4 +1,5 @@
 !.*
+.sl
```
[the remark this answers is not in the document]
altano: It allows me to use [Sapling SCM](https://sapling-scm.com/), an alternative front-end to git. I'll remove it but now I have to use git and be sad =[.

cspell.json
```diff
@@ -24,6 +26,8 @@
 		"packagejson",
 		"quickstart",
 		"ruleset",
+		"substringvalue",
+		"substringvalueeeeee",
```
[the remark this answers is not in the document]
altano: I can remove the other non-words from cspell.json, but I have to add "altano" to the dictionary because it's spell checking package names in package.json

altano:
The compliance action failed with this message:

```
Warning: PR Body did not match required format
Error: Resource not accessible by integration
```

🤷🏽‍♀️

altano:
Thanks for the review. All comments resolved or replied to. Back in your court!

src/tests/rules/valid-repository-directory.test.ts
JoshuaKGoldberg: I _think_ it handles it correctly... but don't have the energy to get my old Surface Laptop back in working order or set up an online repro on this little nit. Works for me 😄

cspell.json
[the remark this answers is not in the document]
JoshuaKGoldberg: ye ole #381 -> https://github.com/streetsidesoftware/cspell/issues/3215

PR #632, opened 2024-11-16
feat: sort alphabetically with co-located hooks for `scripts`

### LINKED ISSUE #499: 🚀 Feature: sort-collections: alphabetically sorting with co-located hooks for `scripts`
### Bug Report Checklist

### Overview

npm `run-scripts` support [`pre` and `post` hooks](https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts). For readability I prefer them to be co-located next to their script they belong to,

e.g.

```
{
  "scripts": {
    "precompress": "",
    "compress": "",
    "postcompress": ""
  }
}
```

while `pre` comes before and `post` after the actual script.

So this should be valid:

```
{
  "scripts": {
    "build": "",
    "precompress": "",
    "compress": "",
    "postcompress": ""
  }
}
```

and this one invalid:

```
{
  "scripts": {
    "precompress": "",
    "compress": "",
    "postcompress": "",
    "build": ""
  }
}
```

### Additional Info

_No response_

sasial-dev:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #499

## Overview

Fixes #499 and adds tests 😄 

I'm not 100% sold on how I did the implementation, and I'm curious to know your thoughts on it.

### DIFF
--- a/src/rules/sort-collections.ts
+++ b/src/rules/sort-collections.ts
@@ -29,6 +29,12 @@ export const rule = createRule<Options>({
 					toSort.includes(key.value)
 				) {
 					const currentOrder = collection.properties;
+					const scripts = new Set(
+						currentOrder.map(
+							(prop) => (prop.key as AST.JSONStringLiteral).value,
+						),
+					);
+
 					const desiredOrder = currentOrder.slice().sort((a, b) => {
 						let aKey = (a.key as AST.JSONStringLiteral).value;
 						let bKey = (b.key as AST.JSONStringLiteral).value;
@@ -38,18 +44,30 @@ export const rule = createRule<Options>({
 						} else {
 							let modifier = 0;
 
-							if (aKey.startsWith("pre")) {
+							if (
+								aKey.startsWith("pre") &&
+								scripts.has(aKey.substring(3))
+							) {
 								aKey = aKey.substring(3);
 								modifier -= 1;
-							} else if (aKey.startsWith("post")) {
+							} else if (
+								aKey.startsWith("post") &&
+								scripts.has(aKey.substring(4))
+							) {
 								aKey = aKey.substring(4);
 								modifier += 1;
 							}
 
-							if (bKey.startsWith("pre")) {
+							if (
+								bKey.startsWith("pre") &&
+								scripts.has(bKey.substring(3))
+							) {
 								bKey = bKey.substring(3);
 								modifier += 1;
-							} else if (bKey.startsWith("post")) {
+							} else if (
+								bKey.startsWith("post") &&
+								scripts.has(bKey.substring(4))
+							) {
 								bKey = bKey.substring(4);
 								modifier -= 1;
 							}
--- a/src/rules/sort-collections.ts
+++ b/src/rules/sort-collections.ts
@@ -29,14 +29,38 @@ export const rule = createRule<Options>({
 					toSort.includes(key.value)
 				) {
 					const currentOrder = collection.properties;
-					const desiredOrder = currentOrder
-						.slice()
-						.sort((a, b) =>
-							(a.key as AST.JSONStringLiteral).value >
-							(b.key as AST.JSONStringLiteral).value
-								? 1
-								: -1,
-						);
+					const desiredOrder = currentOrder.slice().sort((a, b) => {
+						let aKey = (a.key as AST.JSONStringLiteral).value;
+						let bKey = (b.key as AST.JSONStringLiteral).value;
+
+						if (key.value !== "scripts") {
+							return aKey > bKey ? 1 : -1;
+						} else {
+							let modifier = 0;
+
+							if (aKey.startsWith("pre")) {
+								aKey = aKey.substring(3);
+								modifier -= 1;
+							} else if (aKey.startsWith("post")) {
+								aKey = aKey.substring(4);
+								modifier += 1;
+							}
+
+							if (bKey.startsWith("pre")) {
+								bKey = bKey.substring(3);
+								modifier += 1;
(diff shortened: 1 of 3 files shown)


src/rules/sort-collections.ts
```diff
@@ -29,14 +29,38 @@ export const rule = createRule<Options>({
...
-						.slice()
-						.sort((a, b) =>
-							(a.key as AST.JSONStringLiteral).value >
-							(b.key as AST.JSONStringLiteral).value
-								? 1
-								: -1,
-						);
+					const desiredOrder = currentOrder.slice().sort((a, b) => {
+						let aKey = (a.key as AST.JSONStringLiteral).value;
+						let bKey = (b.key as AST.JSONStringLiteral).value;
```
sasial-dev, line 34: Not sure if you prefer this or
```suggestion
						let { value: aKey } = a.key as AST.JSONStringLiteral;
						let { value: bKey } = b.key as AST.JSONStringLiteral;
```

src/rules/sort-collections.ts
```diff
@@ -29,14 +29,38 @@ export const rule = createRule<Options>({
...
+
+						if (key.value !== "scripts") {
+							return aKey > bKey ? 1 : -1;
+						} else {
+							let modifier = 0;
```
sasial-dev, line 39: I'm not sold on how I did this - am curious to know your thoughts

src/rules/sort-collections.ts
JoshuaKGoldberg: Unless a lint rule is yelling at us, I think it's fine as you have it. It's the existing code style and fewer characters.

src/rules/sort-collections.ts
```diff
@@ -29,14 +29,38 @@ export const rule = createRule<Options>({
...
-							(a.key as AST.JSONStringLiteral).value >
-							(b.key as AST.JSONStringLiteral).value
-								? 1
-								: -1,
-						);
+					const desiredOrder = currentOrder.slice().sort((a, b) => {
+						let aKey = (a.key as AST.JSONStringLiteral).value;
+						let bKey = (b.key as AST.JSONStringLiteral).value;
+
+						if (key.value !== "scripts") {
```
JoshuaKGoldberg, line 36: [Refactor] We know whether the `key.value` is `"scripts"` before this function is declared. So instead of checking _inside_ the function:

```ts
.sort((a, b) => {
  if (key.value !== "scripts") {
    // non-scripts logic
  } else {
    // scripts-specific logic
  }
})
```

...I suspect it might be cleaner to pass `.sort()` a different function based on `key.value`:

```ts
.sort(
  key.value === "scripts"
    ? (a, b) => { /* scripts-specific logic */ }
    : (a, b) => { { scripts-specific logic */ }
)
```

I'm not sure. Maybe it'd be super ugly and extra code unnecessarily. What do you think about trying this out and seeing how it looks?

src/rules/sort-collections.ts
JoshuaKGoldberg: Having just one `let modifier` is clever! It seems reasonable if we assume `pre` and `post` always cause these modifiers. But, what happens if the script name just coincidentally happens to have `pre` or `post` and doesn't match other stuff?

```json
{
  "scripts": {
    "lint": "...",
    "poster": "..."
  }
}
```

I would think `"poster"` should come after `"lint"`. Could you add some test coverage for at least the following & make sure they work?

* `"pre"` without a non-prefixed script
* `"post"` without a non-prefixed script
* `"pre" and `"post"` without a non-prefixed script

JoshuaKGoldberg (changes requested):
Great start, sorry for taking so long to get to this! I think you're on the right track - just there's some more edge-case-iness to handle. 🚀

src/rules/sort-collections.ts
sasial-dev: The problem is that `aKey` and `bKey` are used in both. So it'll add duplicate code as `aKey` and `bKey` will be defined twice.

sasial-dev:
@JoshuaKGoldberg - are you ok if I add `postwatch` to CSpell? It seems to be triggering it in the tests

JoshuaKGoldberg:
Yup, that seems like the right way to solve that 👍

sasial-dev:
CI should pass now.

src/tests/rules/sort-collections.test.ts
JoshuaKGoldberg, line 1: [Testing] This is tricky logic and I worry about it breaking over time. I think the following should all exist as `invalid` cases:

* `build`, `postbuild`, and `prebuild` all exist
* `build` and `postbuild` exist, but not `prebuild`
* `build` and `prebuild` exist, but not `postbuild`
* `prebuild` and `prebuild` exist, but not `postbuild`

...and the one with all three should probably be a `valid` case.

JoshuaKGoldberg (changes requested):
Thanks, this is looking good! I took a deeper look and I think this deserves a bit more unit testing care than most other changes.

src/tests/rules/sort-collections.test.ts
sasial-dev: Using the second bullet point as an example, do you mean an invalid case where `build` and `postbuild` exist?

JoshuaKGoldberg: Exactly yeah. As in, they're in order `["postbuild", "build"]` and are fixed to `["build", "postbuild"]` order.

JoshuaKGoldberg (approved):
Melissa Peterman as Bonnie Wheeler joyfully saying "it's glorious"

sasial-dev:
Thank you!

PR #711, opened 2024-12-30
feat(valid-package-definition): rename rule from valid-package-def

### LINKED ISSUE #132: 🚀 Feature: Rename valid-package-def to avoid abbreviation
### Bug Report Checklist

### Overview

I don't like abbreviations in coding. I find it hard to keep track of them. Especially for folk whose first language was not English, they can reduce readability.

The only rule this package contains with an abbreviated name is `valid-package-def`. Let's rename it to ... `valid-package-definition`? `valid-package-schema`? `valid-package-file`? Some bikeshedding

### Additional Info

_No response_

michaelfaith:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #132 

## Overview

This change deprecates the `valid-package-def` rule in favor of a new `valid-package-definition` rule, providing the same behavior.  The deprecated version of the rule will be removed in the next major release.  

I assumed you wanted to roll this change out in a non-breaking way initially?  If that's not the case and you want to major, I can remove the deprecated rule.  If you do prefer this approach, then it may be worth documenting the follow-up so that when it's time to major, the deprecated rule is dropped?

Closes #132


JoshuaKGoldberg (approved):
👍 LGTM, I agree with the strategy of keeping the old rule around as deprecated.

We're on 0.x so technically we can do whatever we want... but as a nicety for users I do like keeping the old one around for a while.

PR #721, opened 2025-01-05
feat(no-redundant-files): add new rule

### LINKED ISSUE #686: 🚀 Feature: Rule to flag redundant entries in package.json "files"
### Bug Report Checklist

### Overview

From https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files:

> Certain files are always included, regardless of settings:
> 
> * `package.json`
> * `README`
> * `LICENSE / LICENCE`
> * `The file in the "main" field`
> * `The file(s) in the "bin" field`

TIL I've been unnecessarily including at least `package.json` in my packages. Amusing.

It'd be nice to have a lint rule that reports on them & autofixes them away.

### Additional Info

Putting as `status: in discussion` for a bit because I suspect this might already exist somewhere else. Maybe one of the equivalents mentioned in https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/42? This feels like it should exist already...

Similar-ish to #683.

💖

michaelfaith:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #686 

## Overview

This change adds a new rule for preventing inclusion of unnecessary or redundant files in a package.json's `files` list.  It checks for two primary types of errors:
1. Duplicate entries within the files array (the same file listed more than once)
1. Files that are automatically included by npm, and don't need to be explicitly included.

Of the second type, there are two flavors that npm includes automatically
1. Files that are always included, regardless of what else is present in the package.json (e.g. README.md)
1. Files that are included because they're declared in other places in the package (e.g. file declared as the `main` entry)

The one thing I wasn't sure about, is whether this new rule should be included in recommended or not (which would technically be a breaking change...)

Closes: #686

### DIFF
--- a/src/rules/no-redundant-files.ts
+++ b/src/rules/no-redundant-files.ts
@@ -1,0 +1,229 @@
+import type { AST as JsonAST } from "jsonc-eslint-parser";
+
+import * as ESTree from "estree";
+
+import { createRule } from "../createRule.js";
+import { isJSONStringLiteral, isNotNullish } from "../utils/predicates.js";
+
+const defaultFiles = [
+	/* cspell:disable-next-line */
+	"LICENCE",
+	/* cspell:disable-next-line */
+	"LICENCE.md",
+	"LICENSE",
+	"LICENSE.md",
+	"package.json",
+	"README.md",
+] as const;
+
+const cachedRegex = new Map<string, RegExp>();
+const getCachedLocalFileRegex = (filename: string) => {
+	// Strip the leading `./`, if there is one, since we'll be incorporating
+	// it into the regex.
+	const baseFilename = filename.replace("./", "");
+	let regex = cachedRegex.get(baseFilename);
+	if (regex) {
+		return regex;
+	} else {
+		regex = new RegExp(`^(./)?${baseFilename}$`, "i");
+		cachedRegex.set(baseFilename, regex);
+		return regex;
+	}
+};
+
+export const rule = createRule({
+	create(context) {
+		// We need to cache these as we find them, since we need to know some of
+		// the other values to ensure that files doesn't contain duplicates.
+		const entryCache: {
+			bin: string[];
+			files: (JsonAST.JSONExpression | null)[];
+			main?: string;
+		} = { bin: [], files: [] };
+
+		/**
+		 * Report rule violations
+		 */
+		const report = (
+			elements: (JsonAST.JSONExpression | null)[],
+			index: number,
+			messageId: string,
+		) => {
+			const element = elements[index];
+
+			if (isNotNullish(element) && isJSONStringLiteral(element)) {
+				context.report({
+					data: { file: element.value },
+					messageId,
+					node: element as unknown as ESTree.Node,
+					suggest: [
+						{
+							*fix(fixer) {
+								yield fixer.remove(
+									element as unknown as ESTree.Node,
+								);
+
+								// If this is not the last entry, then we need to remove the comma from this line.
+								const tokenFromCurrentLine =
+									context.sourceCode.getTokenAfter(
+										element as unknown as ESTree.Node,
+									);
+								if (tokenFromCurrentLine?.value === ",") {
+									yield fixer.remove(tokenFromCurrentLine);
+								}
+
+								// If this is the last line and it's not the only entry, then the line above this one
+								// will become the last line, and should not have a trailing comma.
+								if (
+									index > 0 &&
+									tokenFromCurrentLine?.value !== ","
+								) {
+									const tokenFromPreviousLine =
+										context.sourceCode.getTokenAfter(
+											elements[
+												index - 1
+											] as unknown as ESTree.Node,
+										);
+									if (tokenFromPreviousLine?.value === ",") {
+										yield fixer.remove(
+											tokenFromPreviousLine,
+										);
+									}
+								}
+							},
+							messageId: "remove",
+						},
+					],
+				});
+			}
+		};
+
+		return {
+			"Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.value=bin]"(
+				node: JsonAST.JSONProperty,
+			) {
+				const binValue = node.value;
+
+				// "bin" can be either a simple string or a map of commands to files.
+				// If it's anything else, then this is malformed and we can't really
+				// do anything with it.
+				if (isJSONStringLiteral(binValue)) {
+					entryCache.bin.push(binValue.value);
+				} else if (binValue.type === "JSONObjectExpression") {
+					for (const prop of binValue.properties) {
+						if (isJSONStringLiteral(prop.value)) {
+							entryCache.bin.push(prop.value.value);
+						}
+					}
+				}
+			},
+			"Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.value=files]"(
+				node: JsonAST.JSONProperty,
+			) {
+				// "files" should only ever be an array of strings.
+				if (node.value.type === "JSONArrayExpression") {
+					// We want to add it to the files cache, but also check for
+					// duplicates as we go.
+					const seen = new Set<string>();
+					const elements = node.value.elements;
+					entryCache.files = elements;
+					for (const [index, element] of elements.entries()) {
+						// We only care about JSONStringLiteral values
+						// That _should_ be all that's here, be in order to process
+						// the fix correctly we'll act on the full array of elements
+						if (
+							isNotNullish(element) &&
+							isJSONStringLiteral(element)
+						) {
+							if (seen.has(element.value)) {
+								report(elements, index, "duplicate");
+							} else {
+								seen.add(element.value);
+							}
+
+							// We can also go ahead and check if this matches one
+							// of the static default files
+							const regex = getCachedLocalFileRegex(
+								element.value,
+							);
+							for (const defaultFile of defaultFiles) {
+								if (regex.test(defaultFile)) {
+									report(
+										elements,
+										index,
+										"unnecessaryDefault",
+									);
+								}
+							}
+						}
+					}
+				}
+			},
+			"Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.value=main]"(
+				node: JsonAST.JSONProperty,
+			) {
+				// "main" should only ever be a string.
+				if (isJSONStringLiteral(node.value)) {
+					entryCache.main = node.value.value;
+				}
+			},
+			"Program:exit"() {
+				// Now that we have all of the entries, we can check for unnecessary files.
+				const files = entryCache.files;
+
+				// Bail out early if there are no files.
+				if (files.length === 0) {
+					return;
+				}
+
+				const validations = [
+					// First check if the "main" entry is included in "files".
+					{
+						files: entryCache.main ? [entryCache.main] : [],
+						messageId: "unnecessaryMain",
+					},
+					// Next check if any "bin" entries are included in "files".
+					{
+						files: entryCache.bin,
+						messageId: "unnecessaryBin",
+					},
+				];
+				for (const validation of validations) {
+					for (const fileToCheck of validation.files) {
+						for (const [index, fileEntry] of files.entries()) {
+							if (
+								isNotNullish(fileEntry) &&
+								isJSONStringLiteral(fileEntry)
+							) {
+								const regex = getCachedLocalFileRegex(
+									fileEntry.value,
+								);
+								if (regex.test(fileToCheck)) {
+									report(files, index, validation.messageId);
+								}
+							}
+						}
+					}
+				}
+			},
+		};
+	},
+
+	meta: {
+		docs: {
+			category: "Best Practices",
+			description: "Prevents adding unnecessary / redundant files.",
+			recommended: false,
+		},
+		hasSuggestions: true,
+		messages: {
+			duplicate: 'Files has more than one entry for "{{file}}".',
+			remove: "Remove this redundant entry.",
+			unnecessaryBin: `Explicitly declaring "{{file}}" in "files" is unnecessary; it's included in "bin".`,
+			unnecessaryDefault: `Explicitly declaring "{{file}}" in "files" is unnecessary; it's included by default.`,
+			unnecessaryMain: `Explicitly declaring "{{file}}" in "files" is unnecessary; it's the "main" entry.`,
+		},
+		schema: [],
+		type: "suggestion",
+	},
+});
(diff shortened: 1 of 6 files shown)


src/rules/no-redundant-files.ts
```diff
@@ -0,0 +1,272 @@
...
+					for (const [index, element] of elements.entries()) {
+						if (seen.has(element.value)) {
+							context.report({
+								data: { file: element.value },
+								messageId: "duplicate",
+								node: element as unknown as ESTree.Node,
+								suggest: [
+									{
+										fix: (fixer) =>
+											removeEntryAndAppropriateCommas(
```
JoshuaKGoldberg, line 141: [Refactor] 🤔 The `removeEntryAndAppropriateCommas` call seems to be _almost_ the exact same in all four of the places it's used. Could you please refactor to a shared helper in some way?

I ask because this file is kind of long for a first version of a lint rule, and so I'm having a hard time reading through it. The comments are definitely helpful though!

---

Btw, just playing around with the `Program:exit`, I got all tests to pass with this fun pattern:

```ts
for (const [messageId, entries] of [
    // First check if the "main" entry is included in "files".
    ["unnecessaryMain", [entryCache.main]],

    // Next check if any "bin" entries are included in "files".
    ["unnecessaryBin", entryCache.bin],
] as const) {
    for (const entry of entries) {
        if (entry) {
            for (const [index, fileEntry] of files.entries()) {
								// (roughly the same code as before    
```

No worries if you hate this and don't want it. Just posting in case it's useful. 🙂

src/rules/no-redundant-files.ts
```diff
@@ -0,0 +1,272 @@
...
+	// will become the last line, and should not have a trailing comma.
+	if (currentIndex > 0) {
+		const commaTokenFromPreviousLine =
+			context.sourceCode.getTokenAfter(
+				elements[currentIndex - 1] as unknown as ESTree.Node,
+			)?.value === ","
+				? context.sourceCode.getTokenAfter(
+						elements[currentIndex - 1] as unknown as ESTree.Node,
+					)
+				: null;
```
JoshuaKGoldberg, line 83: [Testing] Looks like this isn't covered by unit tests. Could you please add in test coverage anywhere that needs it?

src/rules/no-redundant-files.ts
```diff
@@ -0,0 +1,272 @@
...
+					element as unknown as ESTree.Node,
+				)
+			: null;
+	if (commaTokenFromCurrentLine) {
+		fixes.push(fixer.remove(commaTokenFromCurrentLine));
+	}
+
+	// If this is the last line and it's not the only entry, then the line above this one
+	// will become the last line, and should not have a trailing comma.
+	if (currentIndex > 0) {
```
JoshuaKGoldberg, line 75: [Refactor] Quick perf improvement: moving the `if (!commaTokenFromCurrentLine` up here means we can skip the `getTokenAfter`:

```suggestion
	if (currentIndex > 0 && !commaTokenFromCurrentLine) {
```

(note: when combining this with the other refactors in this function, a `?.value !== ","` is needed)

src/rules/no-redundant-files.ts
```diff
@@ -0,0 +1,272 @@
...
+ * @returns The list of fixes that the `context.report` fix needs to perform the removals for us
+ */
+const removeEntryAndAppropriateCommas = (
+	elements: JsonAST.JSONStringLiteral[],
+	currentIndex: number,
+	fixer: Rule.RuleFixer,
+	context: Rule.RuleContext,
+): Rule.Fix[] => {
+	const element = elements[currentIndex];
+	const fixes = [fixer.remove(element as unknown as ESTree.Node)];
```
JoshuaKGoldberg, line 58: [Style] WDYT about switching this to a `function*`? It means we can get rid of a few lines of code & complexity with the array.

```ts
function* removeEntryAndAppropriateCommas( ... ) {
```

```suggestion
	yield fixer.remove(element as unknown as ESTree.Node);
```

I'm in favor, but open to being told this is annoying. 😄

JoshuaKGoldberg (changes requested):
Whoohoo, this is awesome! 🥳 Very excited to see this working!

I ended up leaving comments mostly around refactoring for readability. Please push back if you think I'm being too nitpicky. But either way the functionality looks very good to me!

src/rules/no-redundant-files.ts
michaelfaith: Updated

michaelfaith:
Should we disable this rule on lint rules?  I feel like the visitor functions are more intuitive when the exit functions follow the entry functions, with program:exit at the end.  Happy to change it, but it kind of feels wrong haha 
image

src/rules/no-redundant-files.ts
```diff
@@ -0,0 +1,272 @@
...
+	// If this is not the last entry, then we need to remove the comma from this line.
+	// Or if it's jsonc, then we may need to remove the comma, even if it is the last line
+	const commaTokenFromCurrentLine =
+		context.sourceCode.getTokenAfter(element as unknown as ESTree.Node)
+			?.value === ","
+			? context.sourceCode.getTokenAfter(
+					element as unknown as ESTree.Node,
+				)
+			: null;
+	if (commaTokenFromCurrentLine) {
```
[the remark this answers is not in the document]
michaelfaith: Updated

src/rules/no-redundant-files.ts
michaelfaith: Good call.  Updated

src/rules/no-redundant-files.ts
michaelfaith: This line went away with the above refactor.

src/rules/no-redundant-files.ts
michaelfaith: Updated.  I consolidated the two checks in `Program:exit`, since those were nearly identical, as you pointed out.  The other two would not be as clean, and I fear would hurt readability.  There's often a balance between brevity and readability.  I think going any further could tip that balance in the wrong direction.  But happy to hear your feedback on this iteration

src/rules/no-redundant-files.ts
```diff
@@ -0,0 +1,252 @@
...
+	const baseFilename = filename.replace("./", "");
+	let regex = cachedRegex.get(baseFilename);
+	if (regex) {
+		return regex;
+	} else {
+		regex = new RegExp(`^(./)?${baseFilename}$`, "i");
+		cachedRegex.set(baseFilename, regex);
+		return regex;
+	}
+};
```
JoshuaKGoldberg, line 34: [Refactor] I've seen this "make a Map for a single cache" pattern a bunch, I made a [`cached-factory`](https://www.npmjs.com/package/cached-factory) package for it. Up for using it to reduce code here? IMO this is a bit more readable:

```suggestion
import { CachedFactory } from "cached-factory";

const regexFactory = new CachedFactory(
	(baseFilename: string) => new RegExp(`^(./)?${baseFilename}$`, "i"),
);

const getCachedLocalFileRegex = (filename: string) =>
	regexFactory.get(filename.replace("./", ""));
```

(not a blocker at all)

JoshuaKGoldberg:
> Uh oh! @michaelfaith, at least one image you shared is missing helpful alt text. Check https://github.com/[/pull/721](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/721)#issuecomment-2594302487 to fix the following violations:
> * Images should have meaningful alternative text (alt text) at line 2

Aside: interesting bug in the action 🤔. Filed https://github.com/github/accessibility-alt-text-bot/issues/62.

JoshuaKGoldberg:
> Should we disable this rule on lint rules? I feel like the visitor functions are more intuitive when the exit functions follow the entry functions, with program:exit at the end. Happy to change it, but it kind of feels wrong haha

Ha, yeah, it does feel odd. I do have a preference for the perfectionist plugin workin on the other parts of rule files though. Do you see a good way to configure it to always put `:exit` keys last? I haven't dug into its configurations deeply in a while, but suspect it's probably doable.

In the meantime I'd suggest putting an inline ESLint disable comment on the line.

src/rules/no-redundant-files.ts
```diff
@@ -0,0 +1,252 @@
...
+ * we don't need to worry about removing this element's comma, but we should remove
+ * the comma from the element before, since it will the last element after the removal.
+ * If this is not the last line, and it's not the only line (which wouldn't have a comma)
+ * then we need to remove the comma from this line, so that it's not left behind after
+ * the node removal.
+ * @param elements All file entry elements
+ * @param currentIndex The index of the entry that we want to remove
+ * @param fixer The eslint fixer function we'll use to remove the entry
+ * @param context Current context
+ * @yields Fixes that the `context.report` fix needs to perform the removals for us
```
JoshuaKGoldberg, line 49: [Docs] I think these are all implied by the doc comment, the names of the parameters, and how the function is used. But if you really want them I won't block on this. 🙂 

```suggestion
```

src/rules/no-redundant-files.ts
JoshuaKGoldberg: Agreed, I think you found the right balance. Thanks!

JoshuaKGoldberg: I tinkered with this locally and found myself liking a function along the lines of this, placed inside the `create(context) { ... }`:

```ts
function report(
    element: JsonAST.JSONStringLiteral,
    index: number,
    messageId: MessageId,
) {
    const { elements } = element.parent as JsonAST.JSONArrayExpression;

    context.report({
        data: { file: element.value },
        messageId,
        node: element as unknown as ESTree.Node,
        suggest: [
            {
                *fix(fixer) {
                    yield* removeEntryAndAppropriateCommas(
                        elements,
                        index,
                        fixer,
                        context,
                    );
                },
                messageId: "remove",
            },
        ],
    });
}
```

Then this this becomes a one-liner in the places that need to report:

```ts
if (seen.has(element.value)) {
	report(element, index, "duplicate");
} else {
```

```ts
if (regex.test(defaultFile)) {
	report(element, index, "unnecessaryDefault");
}
```

```ts
if (regex.test(fileToCheck)) {
	report(fileEntry, index, validation.messageId);
}
```

Note that `elements` has a type error with `removeEntryAndAppropriateCommas`; the type should include `| null` and can be the more wide `(JSONExpression | null)[]` without issue. I'll comment separately on the null point.

Going one step further, I personally would even inline the `removeEntryAndAppropriateCommas` function into the `*fix`, since the function is now only used in that one place. A bit less of a need to add the big doc comment to it.

I personally find either level a bit clearer to read, and it's about ~two dozen fewer lines in the first step -> ~three dozen lines fewer on the second step after removing most or all doc comments. Do 1-2 of those steps feel reasonable to you?

src/rules/no-redundant-files.ts
michaelfaith: Right on.  Do you think it's worth the additional dependency?

michaelfaith:
> Do you see a good way to configure it to always put :exit keys last?

I put in a change to `Program:exit` last.  And it's certainly possible to put all `:exit` entries after everything else, but what I really want is to put each `:exit` after it's corresponding enter, and unfortunately I don't see a great way to do that with the config.

src/rules/no-redundant-files.ts
```diff
@@ -0,0 +1,252 @@
...
+				node: JsonAST.JSONProperty,
+			) {
+				// "files" should only ever be an array of strings.
+				if (node.value.type === "JSONArrayExpression") {
+					// We want to add it to the files cache, but also check for
+					// duplicates as we go.
+					const seen = new Set<string>();
+					const elements = node.value.elements
+						.filter(isNotNullish)
+						.filter(isJSONStringLiteral);
```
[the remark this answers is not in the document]
michaelfaith: Interesting.  Not a use case I really anticipated.  I'll look into it

src/rules/no-redundant-files.ts
michaelfaith: I guess I'm used to abiding by jsdoc's linting rules, which requires params for everything 😁

src/rules/no-redundant-files.ts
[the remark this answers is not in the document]
michaelfaith: I added a couple of tests with nulls added to the input arrays, including the above example, and they're still passing.  Curious where you're seeing that fail?

```js
{
			code: `{
\t"files": [
\t\tnull,
\t\t"README.md",
\t\t"./package.json"
    ]
}
`,
			errors: [
				{
					data: { file: "README.md" },
					line: 4,
					messageId: "unnecessaryDefault",
					suggestions: [
						{
							messageId: "remove",
							output: `{
\t"files": [
\t\tnull,
\t\t
\t\t"./package.json"
    ]
}
`,
						},
					],
				},
				{
					data: { file: "./package.json" },
					line: 5,
					messageId: "unnecessaryDefault",
					suggestions: [
						{
							messageId: "remove",
							output: `{
\t"files": [
\t\tnull,
\t\t"README.md"
\t\t
    ]
}
`,
						},
					],
				},
			],
		},
```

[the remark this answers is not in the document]
michaelfaith: Ah, nevermind. I see what you mean.  that example is fine, it's this one that will cause it to fail: 😊
```json
{
  "files": [
    "README.md",
    null,
    "./package.json"
  ]
}

```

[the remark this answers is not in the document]
michaelfaith: I'll add a test for that, and address the gap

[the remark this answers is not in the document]
michaelfaith: Updated and resolve

src/rules/no-redundant-files.ts
michaelfaith: Updated.  The only main difference is I flipped it to pass in the full `elements` array and index, rather than individual index and getting the `parent`.  It felt more intuitive that the params are the array along with the index into that array, rather than the JSON String literal element, which may or may not be in an array with an index.  Also I'd have to validate that the parent of this string literal is actually an array type.  So yeah, this just felt a little cleaner and didn't add any more code.

JoshuaKGoldberg: 💡 I like that more, nice!

JoshuaKGoldberg (approved):
🚀 Awesome! This is looking great, and I'm really excited to get it in. Thanks so much for iterating on it a bunch @michaelfaith, you did great work here!

I can go ahead and merge this whenever you're happy with it. Since we have an open thread I'll wait for your decision on what to do about it. If you think it's perfect as-is, great, I can take it from here. 👍

src/rules/no-redundant-files.ts
michaelfaith: Updated to use `cached-factory`

michaelfaith: @JoshuaKGoldberg looks like `cached-factory` is ESM only, which is causing the cjs build to fail.  I'm going to go ahead and revert that change

michaelfaith:
@JoshuaKGoldberg reverted the `cached-factory` change. I believe it's in a good spot, whenever you're ready to merge.

src/rules/no-redundant-files.ts
JoshuaKGoldberg: Ah, right, I forgot about that. Sounds like a wontfix comment to me. 🙂 thanks for the investigation!

JoshuaKGoldberg (approved):
💯

PR #741, opened 2025-01-18
feat: add no-empty-fields rule

### LINKED ISSUE #683: 🚀 Feature: Rule to flag empty arrays/objects that do nothing
### Bug Report Checklist

### Overview

Quite a few package entries do nothing if they're an empty array or object. Some top-of-mind examples:

* `dependencies`/`devDependencies`/`peerDependencies`
* `files`
* `scripts`

It'd be nice to have a lint rule that reports on them & autofixes them away.

### Additional Info

This is similar on the surface to https://github.com/eslint/json/issues/54, in that it's a rule for blank/empty entries. The difference here is that `eslint-plugin-package-json` is for a specific JSON schema. `@eslint/json` is general to all JSON.

Putting as `status: in discussion` for a bit because I suspect this might already exist somewhere else. Maybe one of the equivalents mentioned in #42? This feels like it should exist already...

This issue prompted by seeing a very nice cleanup by @crutchcorn in https://github.com/TryGhost/Ghost/pull/21818/files#r1873314451. Thanks Corbin! 😄 

💖

chouchouji:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #683

### DIFF
--- a/docs/rules/no-empty-fields.md
+++ b/docs/rules/no-empty-fields.md
@@ -1,0 +1,7 @@
+# no-empty-fields
+
+💼 This rule is enabled in the ✅ `recommended` config.
+
+💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
--- a/docs/rules/no-empty-fields.md
+++ b/docs/rules/no-empty-fields.md
@@ -5,3 +5,33 @@
 💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
 
 <!-- end auto-generated rule header -->
+
+This rule flags all empty arrays and objects in a `package.json`,
+as such empty expressions do nothing, and are often the result of a mistake.
+It will report both named properties that are empty, as well as nested arrays and objects
+that are empty.
+
+Example of **incorrect** code for this rule:
+
+```json
+{
+	"main": "lib/index.js",
+	"scripts": {},
+	"files": [],
+	"simple-git-hooks": {
+		"pre-commit": "pnpm exec nano-staged --allow-empty",
+		"preserveUnused": []
+	}
+}
+```
+
+Example of **correct** code for this rule:
+
+```json
+{
+	"main": "lib/index.js",
+	"simple-git-hooks": {
+		"pre-commit": "pnpm exec nano-staged --allow-empty"
+	}
+}
+```
--- a/src/rules/no-empty-fields.ts
+++ b/src/rules/no-empty-fields.ts
@@ -2,52 +2,56 @@ import type { AST as JsonAST } from "jsonc-eslint-parser";
 
 import * as ESTree from "estree";
 
-import { createRule } from "../createRule";
+import { createRule, PackageJsonRuleContext } from "../createRule";
+
+const report = (
+	context: PackageJsonRuleContext,
+	node: JsonAST.JSONProperty,
+) => {
+	context.report({
+		data: { field: (node.key as JsonAST.JSONStringLiteral).value },
+		messageId: "emptyFields",
+		node: node as unknown as ESTree.Node,
+		suggest: [
+			{
+				*fix(fixer) {
+					yield fixer.remove(node as unknown as ESTree.Node);
+
+					const tokenFromCurrentLine =
+						context.sourceCode.getTokenAfter(
+							node as unknown as ESTree.Node,
+						);
+					const tokenFromPreviousLine =
+						context.sourceCode.getTokenBefore(
+							node as unknown as ESTree.Node,
+						);
+
+					if (
+						tokenFromPreviousLine?.value === "," &&
+						tokenFromCurrentLine?.value !== ","
+					) {
+						yield fixer.remove(tokenFromPreviousLine);
+					}
+
+					if (tokenFromCurrentLine?.value === ",") {
+						yield fixer.remove(tokenFromCurrentLine);
+					}
+				},
+				messageId: "remove",
+			},
+		],
+	});
+};
 
 export const rule = createRule({
 	create(context) {
-		const report = (node: JsonAST.JSONProperty) => {
-			context.report({
-				data: { field: (node.key as JsonAST.JSONStringLiteral).value },
-				messageId: "emptyFields",
-				node: node as unknown as ESTree.Node,
-				suggest: [
-					{
-						*fix(fixer) {
-							yield fixer.remove(node as unknown as ESTree.Node);
-
-							const tokenFromCurrentLine =
-								context.sourceCode.getTokenAfter(
-									node as unknown as ESTree.Node,
-								);
-							const tokenFromPreviousLine =
-								context.sourceCode.getTokenBefore(
-									node as unknown as ESTree.Node,
-								);
-
-							if (
-								tokenFromPreviousLine?.value === "," &&
-								tokenFromCurrentLine?.value !== ","
-							) {
-								yield fixer.remove(tokenFromPreviousLine);
-							}
-
-							if (tokenFromCurrentLine?.value === ",") {
-								yield fixer.remove(tokenFromCurrentLine);
-							}
-						},
-						messageId: "remove",
-					},
-				],
-			});
-		};
 		return {
 			JSONArrayExpression(node: JsonAST.JSONArrayExpression) {
 				if (
 					!node.elements.length &&
 					node.parent.type === "JSONProperty"
 				) {
-					report(node.parent);
+					report(context, node.parent);
 				}
 			},
 			JSONObjectExpression(node: JsonAST.JSONObjectExpression) {
@@ -55,7 +59,7 @@ export const rule = createRule({
 					!node.properties.length &&
 					node.parent.type === "JSONProperty"
 				) {
-					report(node.parent);
+					report(context, node.parent);
 				}
 			},
 		};
--- a/src/rules/no-empty-fields.ts
+++ b/src/rules/no-empty-fields.ts
@@ -1,7 +1,7 @@
 import type { AST as JsonAST } from "jsonc-eslint-parser";
 
 import { createRule } from "../createRule";
-import { isString } from "../utils/predicates";
+import { isString, omit } from "../utils/predicates";
 
 export const rule = createRule({
 	create(context) {
@@ -14,25 +14,15 @@ export const rule = createRule({
 		const arrayFields = ["files"];
 
 		return {
-			"Program > JSONExpressionStatement > JSONObjectExpression"(
-				node: JsonAST.JSONObjectExpression,
-			) {
-				function getRange(
-					properties: JsonAST.JSONProperty[],
-					property: JsonAST.JSONProperty,
-					index: number,
-				): [number, number] {
-					const isLastProperty = properties.length - 1 === index;
-					// if the property is last, we should remove ',' before this property
-					const start = isLastProperty
-						? properties.slice(-2)[0].range[1]
-						: property.range[0];
-					// if the property isn't last, we should remove ',' after this property
-					const end = property.range[1] + (isLastProperty ? 0 : 1);
-					return [start, end];
-				}
+			"Program:exit"() {
+				const { ast, text } = context.sourceCode;
 
-				node.properties.forEach((property, index) => {
+				const json = JSON.parse(text) as Record<string, unknown>;
+				const { properties } = ast.body[0].expression;
+
+				for (let i = 0; i < properties.length; i += 1) {
+					const property = properties[i];
+
 					if (
 						property.key.type === "JSONLiteral" &&
 						isString(property.key.value) &&
@@ -40,16 +30,22 @@ export const rule = createRule({
 						property.value.type === "JSONObjectExpression" &&
 						!property.value.properties.length
 					) {
+						const field = property.key.value;
 						context.report({
 							data: {
-								property: property.key.value,
+								field,
 							},
 							fix(fixer) {
-								return fixer.removeRange(
-									getRange(node.properties, property, index),
+								return fixer.replaceText(
+									ast,
+									JSON.stringify(
+										omit(json, [field]),
+										null,
+										2,
+									),
 								);
 							},
-							loc: property.loc,
+							loc: properties[i].loc,
 							messageId: "emptyFields",
 						});
 					} else if (
@@ -59,20 +55,26 @@ export const rule = createRule({
 						property.value.type === "JSONArrayExpression" &&
 						!property.value.elements.length
 					) {
+						const field = property.key.value;
 						context.report({
 							data: {
-								property: property.key.value,
+								field,
 							},
 							fix(fixer) {
-								return fixer.removeRange(
-									getRange(node.properties, property, index),
+								return fixer.replaceText(
+									ast,
+									JSON.stringify(
+										omit(json, [field]),
+										null,
+										2,
+									),
 								);
 							},
-							loc: property.loc,
+							loc: properties[i].loc,
 							messageId: "emptyFields",
 						});
 					}
-				});
+				}
 			},
 		};
 	},
@@ -84,7 +86,7 @@ export const rule = createRule({
 		},
 		hasSuggestions: true,
 		messages: {
-			emptyFields: 'Should remove empty "{{property}}"',
+			emptyFields: 'Should remove empty "{{field}}"',
 		},
 		fixable: "whitespace",
 		schema: [],
--- a/src/rules/no-empty-fields.ts
+++ b/src/rules/no-empty-fields.ts
@@ -1,74 +1,57 @@
-import { createRule } from "../createRule";
 import type { AST as JsonAST } from "jsonc-eslint-parser";
 
 import * as ESTree from "estree";
 
+import { createRule } from "../createRule";
+
 export const rule = createRule({
 	create(context) {
+		const report = (
+			node: JsonAST.JSONProperty & { key: JsonAST.JSONStringLiteral },
+		) => {
+			context.report({
+				data: { field: node.key.value },
+				messageId: "emptyFields",
+				node: node as unknown as ESTree.Node,
+				suggest: [
+					{
+						*fix(fixer) {
+							yield fixer.remove(node as unknown as ESTree.Node);
+
+							const tokenFromCurrentLine =
+								context.sourceCode.getTokenAfter(
+									node as unknown as ESTree.Node,
+								);
+							const tokenFromPreviousLine =
+								context.sourceCode.getTokenBefore(
+									node as unknown as ESTree.Node,
+								);
+
+							if (
+								tokenFromPreviousLine?.value === "," &&
+								tokenFromCurrentLine?.value !== ","
+							) {
+								yield fixer.remove(tokenFromPreviousLine);
+							}
+
+							if (tokenFromCurrentLine?.value === ",") {
+								yield fixer.remove(tokenFromCurrentLine);
+							}
+						},
+						messageId: "remove",
+					},
+				],
+			});
+		};
 		return {
-			"Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.type=JSONLiteral]"(
-				node: JsonAST.JSONProperty & {
-					key: JsonAST.JSONStringLiteral;
-				},
-			) {
-				if (
-					!["JSONArrayExpression", "JSONObjectExpression"].includes(
-						node.value.type,
-					)
-				) {
-					return;
+			JSONArrayExpression(node: JsonAST.JSONArrayExpression) {
+				if (!node.elements.length) {
+					report(node.parent as any);
 				}
-
-				if (
-					(node.value.type === "JSONArrayExpression" &&
-						!node.value.elements.length) ||
-					(node.value.type === "JSONObjectExpression" &&
-						!node.value.properties.length)
-				) {
-					context.report({
-						data: { field: node.key.value },
-						messageId: "emptyFields",
-						node: node as unknown as ESTree.Node,
-						suggest: [
-							{
-								*fix(fixer) {
-									const keys = Object.keys(
-										JSON.parse(context.sourceCode.text),
-									);
-									const index = keys.findIndex(
-										(key) => key === node.key.value,
-									);
-									const tokenFromPreviousLine =
-										context.sourceCode.getTokenAfter(
-											node as unknown as ESTree.Node,
-										);
-									if (
-										tokenFromPreviousLine?.value === "," &&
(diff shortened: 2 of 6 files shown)


chouchouji:
@JoshuaKGoldberg I completed code and test cases. Maybe the code should be optimized. Please review it when you have time.

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,105 @@
...
+						const newline = detectNewline.graceful(text);
+						let result =
+							JSON.stringify(
+								omit(json, [field]),
+								null,
+								type === "tab" ? "\t" : indent,
+							) + endCharacters;
+						if (newline === "\r\n") {
+							result = result.replace(/\n/g, newline);
+						}
```
JoshuaKGoldberg, line 75: [Refactor] The `result` variable is only used inside the `fix` function. It's a little confusing to see that it's declared outside the `context.report`. Let's move any variables only used for fixing inside the `fix`.

💡 If you're finding that code needs to be duplicated between the two fixers, you can always write a utility function to deduplicate it. Something like `function report(node) {` is used in other rules such as `no-redundant-files` and `unique-dependencies`.

src/utils/predicates.ts
```diff
@@ -11,3 +11,29 @@ export function isNotNullish<T extends NonNullable<unknown>>(
 ): value is T {
 	return value !== null && value !== undefined;
 }
+
+export function isString(value: unknown): value is string {
+	return typeof value === "string";
+}
```
JoshuaKGoldberg, line 17: [Refactor] I don't think we need such a straightforward function. Just a `typeof` in-place is code is fine.

```suggestion
```

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export const rule = createRule({
+	create(context) {
+		const objectFields = [
+			"peerDependencies",
+			"scripts",
+			"dependencies",
+			"devDependencies",
+		];
+		const arrayFields = ["files"];
```
JoshuaKGoldberg, line 14: [Refactor] There's no need for these to be re-created inside the `create(context) {` function each time. They should be moved to outside the `rule` definition.

JoshuaKGoldberg, line 14: [Feature] These five fields are the examples I listed in #683 but they're certainly not the only ones! Scrolling through https://docs.npmjs.com/cli/configuring-npm/package-json I'm seeing a lot of other fields: `"directories"`, `"config"`, `"bundleDependencies"`, ...

That list of first-party known fields doesn't include ad-hoc fields added by other tools. For example, [`lint-staged`](https://github.com/lint-staged/lint-staged) asks for a `"lint-staged"` object. If that object is empty then as a user I would expect this rule to report.

```jsonc
{
  // ...
  "lint-staged": {}, // <-- unnecessary, should be reported
  // ...
}
```

Feature request: remove these fields. Let's report on _any_ empty array or object. 

If someone has a tool-specific use case for an empty field, then they can always file a feature request for the rule to add an allowlist or something like that.

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,105 @@
...
+							) + endCharacters;
+						if (newline === "\r\n") {
+							result = result.replace(/\n/g, newline);
+						}
+						context.report({
+							data: {
+								property: field,
+							},
+							fix(fixer) {
+								return fixer.replaceText(ast, result);
```
JoshuaKGoldberg, line 51: [Bug] This replaces the _entire_ `package.json` with the stringified result. Some older rules in this repo do that but it's not a great user experience:

* The whole file gets re-formatted, which is inconvenient to see in general and especially inconvenient if you don't also use a formatter
* It slows linting down if multiple rules have fixes because they can't both be applied

See how other rules perform `fixer.remove` on individual nodes. `no-redundant-files` has some nicely commented code in its reports.

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,105 @@
...
+			"devDependencies",
+		];
+		const arrayFields = ["files"];
+
+		return {
+			"Program:exit"() {
+				const { ast, text } = context.sourceCode;
+
+				const json = JSON.parse(text) as Record<string, unknown>;
+				const { properties } = ast.body[0].expression;
```
JoshuaKGoldberg, line 21: [Feature] Continuing the feature request on finding any arbitrary field, this logic only captures root-level properties. I think it should also flag nested properties:

```jsonc
{
  // ...
  "my-tool-stuff": {
    "things": [] // <-- unnecessary, should be reported
  },
  // ...
}
```

Instead of using `"Program:exit"` as the ESQuery selector, I think you'll want to use `JSONArrayExpression` and `JSONObjectExpression` to flag individual nodes.

JoshuaKGoldberg (changes requested):
This is a solid start, thanks for working on it @chouchouji! 🚀 

I left some comments on making it more comprehensive in feature reporting and more targeted in how it fixes code. Please let me know if anything is unclear. 🙂

src/rules/no-empty-fields.ts
JoshuaKGoldberg, line 1: [Testing] Once you're done refactoring, you'll want to make sure unit test coverage is comprehensive. After the suggested feature requests & refactors are done I suspect you won't need any handling around newlines anymore.

💡 Tip: `pnpm run test run --coverage` locally creates a `coverage/index.html`.

JoshuaKGoldberg:
By the way:

> * [x]  Addresses an existing open issue: fixes #000
> * [x]  That issue was marked as [`status: accepting prs`](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
> * [ ]  Steps in [CONTRIBUTING.md](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/main/.github/CONTRIBUTING.md) were taken
> 
> ## Overview
> ## Issue
> close #683

It's customary to put a message like `closes #...` or `fixes #...` inside the checklist. Like:

> * [x]  Addresses an existing open issue: close #683

I edited that in just now. You can read more here: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword. Cheers! 🙂

src/rules/no-empty-fields.ts
chouchouji: I get it. Thanks.

src/rules/no-empty-fields.ts
michaelfaith: @chouchouji this might be a helpful reference for inspiration: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/721  There's some logic in there around dealing with trailing commas on `JSONArrayExpression`s.  It's a similar concept for `.properties` on `JSONObjectExpression`.

chouchouji: > @chouchouji this might be a helpful reference for inspiration: #721 There's some logic in there around dealing with trailing commas on `JSONArrayExpression`s. It's a similar concept for `.properties` on `JSONObjectExpression`.

I see, Thanks.

michaelfaith: @chouchouji Nice work on the refactor!  This is so much closer.  There is still the use-case of nested objects and arrays that this implementation doesn't cover.  Like the example that @JoshuaKGoldberg at the top of this thread.  I just added  test to confirm that has this as the input code, which should report an error for field `nin`:

```js
{
    "name": "test",
    "peerDependencyMeta": {
        "@altano/repository-tools": {
            "optional": true
        },
        "nin": {}
    }
}
```

But it's not currently reporting.  You should be able to achieve this by having a visitor function for both `JSONArrayExpression` and `JSONObjectExpression`, without this prefix `Program > JSONExpressionStatement > JSONObjectExpression > `  This is what's anchoring you to only matching top-level fields.  In order to not duplicate a bunch of code between the two, you could create a function for the `report` code, which is what 80% of them are.  Then the visitor functions just become the if statement about element or properties length.  Does that make sense?  

Pseudo code:
```js
{
  JSONArrayExpression(node: JsonAST.JSONArrayExpression) {
    if (!node.value.elements.length) {
      report(...);  
    }
  },
  JSONObjectExpression(node: JsonAST.JSONObjectExpression) {
    if (!node.value.properties.length) {
      report(...);  
    }
  }
}
```

And really, you could absorb that if statement into the reusable function and just pass the appropriate array into it.  Let me know what you think of that approach.

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,91 @@
...
+						!node.value.properties.length)
+				) {
+					context.report({
+						data: { field: node.key.value },
+						messageId: "emptyFields",
+						node: node as unknown as ESTree.Node,
+						suggest: [
+							{
+								*fix(fixer) {
+									const keys = Object.keys(
```
michaelfaith, line 35: There shouldn't be a need to parse json here.  We already have the full AST available to us.  I see that you're parsing it so that you can get the index, which you then use to determine whether the comma from the previous line should be removed.  Rather than using the index to make that decision you could check if the tokenBefore is a comma (using `sourceCode.getTokenBefore`).  So the condition becomes "if the token before this node is a comma (i.e. it's not the first element) and the token after this node _is not_ a comma (i.e. it's the last element)", and that should be enough to determine whether to remove the tokenBefore.

michaelfaith (commented):
Nice progress!  I just made an additional observation about the `JSON.parse`

src/rules/no-empty-fields.ts
chouchouji: > @chouchouji Nice work on the refactor! This is so much closer. There is still the use-case of nested objects and arrays that this implementation doesn't cover. Like the example that @JoshuaKGoldberg at the top of this thread. I just added test to confirm that has this as the input code, which should report an error for field `nin`:
> 
> ```js
> {
>     "name": "test",
>     "peerDependencyMeta": {
>         "@altano/repository-tools": {
>             "optional": true
>         },
>         "nin": {}
>     }
> }
> ```
> 
> But it's not currently reporting. You should be able to achieve this by having a visitor function for both `JSONArrayExpression` and `JSONObjectExpression`, without this prefix `Program > JSONExpressionStatement > JSONObjectExpression > ` This is what's anchoring you to only matching top-level fields. In order to not duplicate a bunch of code between the two, you could create a function for the `report` code, which is what 80% of them are. Then the visitor functions just become the if statement about element or properties length. Does that make sense?
> 
> Pseudo code:
> 
> ```js
> {
>   JSONArrayExpression(node: JsonAST.JSONArrayExpression) {
>     if (!node.value.elements.length) {
>       report(...);  
>     }
>   },
>   JSONObjectExpression(node: JsonAST.JSONObjectExpression) {
>     if (!node.value.properties.length) {
>       report(...);  
>     }
>   }
> }
> ```
> 
> And really, you could absorb that if statement into the reusable function and just pass the appropriate array into it. Let me know what you think of that approach.

Thanks for your reply. I am a little busy these days. So I spend time making sure that my direction is right first. I get you word.

chouchouji:
@michaelfaith @JoshuaKGoldberg Thanks,  I had update code. It seems that I am walking in a right direction. Please check this video. 

https://github.com/user-attachments/assets/45f5d8c2-f213-4601-b63f-25937ea38a91

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,77 @@
+import type { AST as JsonAST } from "jsonc-eslint-parser";
+
+import * as ESTree from "estree";
+
+import { createRule } from "../createRule";
+
+export const rule = createRule({
+	create(context) {
+		const report = (node: JsonAST.JSONProperty) => {
```
michaelfaith, line 9: Nit: I'd suggest pulling this report function outside of the rule and passing the `context` in, just to avoid it recreating this function every execution.

michaelfaith (commented):
Much cleaner!  Looks great.  Left one minor (non-blocking) nit.  Also, want to give @JoshuaKGoldberg a chance to take another look before calling it done.  I can take care of creating the doc page that CI is blocking for, unless you'd like to take a stab at that too.

src/rules/no-empty-fields.ts
chouchouji: Thanks, I had done it.

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,81 @@
...
+	});
+};
+
+export const rule = createRule({
+	create(context) {
+		return {
+			JSONArrayExpression(node: JsonAST.JSONArrayExpression) {
+				if (
+					!node.elements.length &&
+					node.parent.type === "JSONProperty"
```
JoshuaKGoldberg, line 52: [Testing] Removing the `&& node.parent.type === "JSONProperty"` logic doesn't fail any tests. So one of the following is true:

* The logic is necessary (`node.parent` doesn't always have `.type === "JSONProperty"`), and test cases should be added to exercise this logic
* The logic is unnecessary (`node.parent` always has `.type === "JSONProperty"`), so should be refactored to be unnecessary, removed, or replaced with a type assertion

This brings up a couple of interesting test cases actually:

```jsonc
// package.json
{
  "some-custom-array": [ [] ],
  "some-custom-object": [ {} ],
}
```

What should the rule do in those cases? 

My gut instinct is that it should report on them by default, because that's just a weird thing to do - even in custom fields. Similar to https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/741#discussion_r1921553305: if someone has a tool-specific use case for an empty field, then they can always file a feature request for the rule to add an allowlist or something like that.

cc @michaelfaith

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,81 @@
...
+				) {
+					report(context, node.parent);
+				}
+			},
+		};
+	},
+	meta: {
+		docs: {
+			category: "Best Practices",
+			description: "Remove empty fields",
```
JoshuaKGoldberg, line 70: [Docs] To be a bit more informative...

```suggestion
			description: "Reports on unnecessary empty arrays and objects.",
```

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,81 @@
...
+			recommended: true,
+		},
+		hasSuggestions: true,
+		messages: {
+			emptyFields: 'Should remove empty "{{field}}"',
```
JoshuaKGoldberg, line 75: [Docs]

```suggestion
			emptyFields: 'This empty "{{field}}" does nothing and can be removed.',
```

docs/rules/no-empty-fields.md
JoshuaKGoldberg, line 1: [Docs] Could you please write up descriptions and examples for this file, similar to other rule docs files?

JoshuaKGoldberg (changes requested):
Progress 🚀

docs/rules/no-empty-fields.md
chouchouji: I think I can have a try.

src/rules/no-empty-fields.ts
chouchouji: I think they are two cases.

1. It includes key and value, e.g.   `"xxx": [], "yyy": {}`. The array or object have one key
2. It includes value only, e.g.  `"xxx": [ {}, [] ]`. The array or object have no key

So I should update my code like this code:

```js
JSONArrayExpression(node: JsonAST.JSONArrayExpression) {
    if (!node.elements.length) {
	report(
	   context,
	  node.parent.type === "JSONProperty" ? node.parent : node,
	);
     }
},
```

You can check the video


https://github.com/user-attachments/assets/4979b02e-4c15-4e20-838b-d0ceaf9a6fca

michaelfaith: Hmm, that's a good question.  I do wonder if there might be a legitimate use case for this.  Such as resetting some tool config?  I.e. some tool has a default config that's applied, in a root level package.json, and a nested package.json wants to clear the config at the lower level?  Not sure how likely that is, but it seems like something we may not want to be too heavy handed with?

docs/rules/no-empty-fields.md
michaelfaith: If they're not comfortable writing this due to any kind of language barrier (or just not being super confident with documentation writing), I don't mind taking that up myself.

chouchouji: > @chouchouji if you end up having any difficulty, I'm happy to help with it.

Thanks, I will write docs after completing this rule logic.

src/rules/no-empty-fields.ts
michaelfaith: In this first iteration, I'd advocate for keeping it simple.  And if we can't articulate a substantive real-world use-case, then I think @JoshuaKGoldberg 's suggestion of covering it for now, until someone brings a valid usecase forward makes sense.  It's easier / less intrusive to relax a rule than it is to make a rule more strict. So, if you already have the empty nested case working, let's go with that.

chouchouji: > In this first iteration, I'd advocate for keeping it simple. And if we can't articulate a substantive real-world use-case, then I think @JoshuaKGoldberg 's suggestion of covering it for now, until someone brings a valid usecase forward makes sense. It's easier / less intrusive to relax a rule than it is to make a rule more strict. So, if you already have the empty nested case working, let's go with that.

I agree with you. Can you review this latest code? I have the empty nested case working. Please tell me if something need to improve.

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,84 @@
...
+	},
+	meta: {
+		docs: {
+			category: "Best Practices",
+			description: "Reports on unnecessary empty arrays and objects.",
+			recommended: true,
+		},
+		hasSuggestions: true,
+		messages: {
+			emptyFields:
```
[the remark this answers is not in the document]
michaelfaith: Oh, and I'm not sure if you're using any kind of AST explorer, but I've found this helpful when working with the jsonc parser: https://ota-meshi.github.io/jsonc-eslint-parser/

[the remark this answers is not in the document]
chouchouji: > Yep, good observation. We should use this message when what we're removing is a property with a name, and a different version of the message when it's an unnamed array/object, which won't have a field name to pass in.
> 
> Also, would you mind adding a test similar to this, just to ensure nested empty elements, in an array that has an element we should preserve are handled correctly. Your current logic should work with this, but it would be good to include
> 
> ```json
> {
>   "name": "test",
>   "config": [[], ["test"]]
> }
> ```

 I am not sure how to update this message. This test case looks like good, I will add it.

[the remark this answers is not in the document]
chouchouji: @michaelfaith I just remove `field` first. If we want to update this message, I will modify it laster. Please check the secondary test case.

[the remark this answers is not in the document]
michaelfaith: I would suggest having two versions of the message:
```js
messages: {
  emptyFields: "The field '{{field}}' does nothing and can be removed.",
  emptyExpression: "This {{expressionType}} does nothing and can be removed.", 
}
```

And if we're removing a `JSONProperty` we use the first message with `data.field` being the name of the property.  If we're removing a `JSONArrayExpression` or `JSONObjectExpression` we use the second one and `data.expressionType` is either 'array' or 'object'.  Providing additional context like that, can help when people are linting from the cli against their entire codebase or from CI, where their errors aren't clickable.  What do you think?

chouchouji:
@JoshuaKGoldberg @michaelfaith Thanks for your help. I had updated code. Please check this video.


https://github.com/user-attachments/assets/e8919843-cd70-4caa-9d7e-d761d5ff4ffc

michaelfaith (commented):
Nice work on the refactor.  The rule logic is looking good to me.  I think fleshing out the documentation page might be all that's left, which I'm happy to help with, if you want.

chouchouji:
> Nice work on the refactor. The rule logic is looking good to me. I think fleshing out the documentation page might be all that's left, which I'm happy to help with, if you want.

Thanks. I have no experience about docs. It would be best if you could write it.

docs/rules/no-empty-fields.md
```diff
@@ -0,0 +1,7 @@
+# no-empty-fields
+
+💼 This rule is enabled in the ✅ `recommended` config.
+
+💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
```
michaelfaith, line 7: Suggested doc page (you'll have to remove the escape character I had to add for it to render on GH correctly).

```suggestion
<!-- end auto-generated rule header -->

This rule flags all empty arrays and objects in a `package.json`,
as such empty expressions do nothing, and are often the result of a mistake.
It will report both named properties that are empty, as well as nested arrays and objects
that are empty.

Example of **incorrect** code for this rule:

```json
{
	"main": "lib/index.js",
	"scripts": {}
}
\```

Example of **correct** code for this rule:

```json
{
	"main": "lib/index.js"
}
\```

```

michaelfaith (commented):
Here's a recommendation for the rule documentation.  What do you think?

chouchouji:
> Here's a recommendation for the rule documentation. What do you think?

Thanks. I think it is very good. I had updated docs. Please review it if you have time.

michaelfaith (commented):
Great.  Nice work on this.  I'll let @JoshuaKGoldberg give it another once over, but I think it's in a good spot.  

One thing I'll call out is that technically this change is a breaking one, since we're adding a new rule to the recommended config.  With the plugin being pre-1.0 and still forming its initial public api, I don't think that's a problem, but just mentioning it for general awareness.

JoshuaKGoldberg:
Yeah the joy of being on a 0.x major is that semver explicitly allows us to add these "breaking" changes whenever we want 🚀. https://semver.org/#spec-item-4:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Obviously we're not going to go _too_ wild with it. I personally try to keep to minor versions, rather than patch versions, for things that would be breaking in a >=1.x.

JoshuaKGoldberg (approved):
💯 this is looking great, thanks!

src/rules/no-empty-fields.ts
```diff
@@ -0,0 +1,122 @@
...
+					if (
+						tokenFromPreviousLine?.value === "," &&
+						tokenFromCurrentLine?.value !== ","
+					) {
+						yield fixer.remove(tokenFromPreviousLine);
+					}
+
+					if (tokenFromCurrentLine?.value === ",") {
+						yield fixer.remove(tokenFromCurrentLine);
+					}
```
JoshuaKGoldberg, line 76: [Non-Actionable] This logic is another example of something that'll be simplified by #750. I'll go ahead and merge this PR now and factor the changes into #750.

PR #750, opened 2025-01-19
chore: use `eslint-fix-utils` for element and property removals

### LINKED ISSUE #747: 🛠 Tooling: Extract element/property removal to a shared package
### Bug Report Checklist

### Overview

I had a 💡 reviewing #721 and #741 just now: the _"remove an element from an array"_ logic is something I've seen before in multiple packages over the years. That and _"remove a property from an object'_ as well. They can actually be tricky to implement because of trailing commas and the like. It'd be nice to have a shared utility package for them.

### Additional Info

I started working on this in https://github.com/JoshuaKGoldberg/eslint-fix-utils. I'll send a draft PR for reference so we can get a feel for it.

💖

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #747

## Overview

Uses the exported APIs added in https://github.com/JoshuaKGoldberg/eslint-fix-utils/commit/6bf5c3329eb47c43d38128328fd4501277b3c8ad.

@michaelfaith I'd love to get your input on this: how does the API feel to you? If you think they should be changed drastically -here and/or in the other repo- I'm definitely up for iterating on them. 🙂 

~~The build failures can be ignored, they're some problem in `eslint-fix-utils` that I haven't dug into yet.~~ ~~Blocked on: https://github.com/JoshuaKGoldberg/eslint-fix-utils/issues/7~~ ✅ `eslint-fix-utils@0.2.0` should work now.

💖

### DIFF
--- a/src/rules/no-empty-fields.ts
+++ b/src/rules/no-empty-fields.ts
@@ -1,5 +1,9 @@
 import type { AST as JsonAST } from "jsonc-eslint-parser";
 
+import {
+	fixRemoveArrayElement,
+	fixRemoveObjectProperty,
+} from "eslint-fix-utils";
 import * as ESTree from "estree";
 
 import { createRule, PackageJsonRuleContext } from "../createRule";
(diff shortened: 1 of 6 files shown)


michaelfaith:
It definitely saved a fair amount of code in the cast of `no-redundant-files`, but it didn't really do anything to help us in `unique-dependencies`.  It prevented us from taking advantage of the fact that you can assume the redundant dependency will not be the last element, and thus will always have a comma token after.  So it ends up being a wash, in terms of value for that one.

As far as the api surface itself, it seems reasonable from the usage side.

JoshuaKGoldberg:
Cool thanks! I'll go ahead and push this all through then. Agreed it's a little unfortunate to lose some of the nice optimizations in `unique-dependencies`. IME the cost of a few token lookups is never really relevant, so I don't mind too much.

src/rules/no-empty-fields.ts
```diff
@@ -52,29 +56,17 @@ const report = (
...
-						tokenFromCurrentLine?.value !== ","
-					) {
-						yield fixer.remove(tokenFromPreviousLine);
-					}
-
-					if (tokenFromCurrentLine?.value === ",") {
-						yield fixer.remove(tokenFromCurrentLine);
-					}
-				},
+				fix:
```
michaelfaith, line 59: Nice to be able also use this on the new rule too!

michaelfaith (approved):
Looks good.  Nice to see the cleanup it allowed for.

Issue #52, opened 2023-09-20
Bring in string value "format-*"/"*-format" rule(s)


JoshuaKGoldberg:
https://npmpackagejsonlint.org/docs/rules#format-rules has three string formatting rules that provide value:

* [description-format](https://npmpackagejsonlint.org/docs/rules/format/description-format)
* [name-format](https://npmpackagejsonlint.org/docs/rules/format/name-format)
* [version-format](https://npmpackagejsonlint.org/docs/rules/format/version-format)

It'd be good to have them in this plugin. But...

* Should they be combined into one rule with options? Split into separate rules?
* If they're separate rules, should the name template be `format-*` (to put together alphabetically) or `*-format` (to associate with the individual property)?
* Note that these use the word "format" to refer to the contents of string property values - _not_ a formatter like Prettier. Is there an equivalent other word that doesn't have the terminology overlap?

Blocked on #40, but once that PR is merged this will be good to go.

JoshuaKGoldberg:
Now that I split out #130 and #131, all that's really left is `description`. Which makes this a much more specific rule! The same question before on _"what do we call this?"_ applies.

JoshuaKGoldberg:
👍 agreed

Issue #51, opened 2023-09-20
Bring in a "require-*"/"no-*" kind of rule


JoshuaKGoldberg:
https://npmpackagejsonlint.org/docs/rules#require-node-rules has a bunch of `require-*` rules for requiring properties to exist, such as [`require-author`](https://npmpackagejsonlint.org/docs/rules/required-node/require-author), [`require-bin`](https://npmpackagejsonlint.org/docs/rules/required-node/require-bin).

This seems like it'd do well as a general-purpose ESLint rule. It could:

* By default: enforce the required properties that will prompt a warning if missing
* By configuration: be extended to any arbitrary property

Maybe, `package-json/require-properties` or `package-json/require-package-properties` as a name? The latter seems a little overly specific, so I'm leaning towards `require-properties` maybe?

But, what about https://npmpackagejsonlint.org/docs/rules/#disallowed-node-rules? Should we unify into one big rule for requiring which properties do or don't exist?

I also think that https://npmpackagejsonlint.org/docs/rules/required-node/require-scripts can be treated as a subset of https://npmpackagejsonlint.org/docs/rules/scripts/prefer-scripts for this rule. As in, this same rule can require both that `scripts` exists _and_ what values it contains. Someone please yell at me if that's not ideal.

~Blocked on #40, but once that PR is merged this will be good to go.~ ✅

JoshuaKGoldberg:
Yup! You'd have to configure the list yourself, as what constitutes as "useless" changes depending on the context.

michaelfaith:
Maybe one rule called `restrict-properties`?  And it has an array for `require` and `disallow`?  Something like that?

michaelfaith:
What criteria do we use for whether a property should have its own rule vs be configurable in the single umbrella rule?  It feels a little odd to have some properties be broken out as dedicated rules while others are configured as part of one big rule.   The importance of a property is such a relative thing, where to draw that line becomes challenging.

JoshuaKGoldberg:
Agreed, having some be granular and some not would be confusing. Proposal: how about we put _all_ properties in their own rules? That's my preference more and more.

michaelfaith:
> Proposal: how about we put _all_ properties in their own rules? That's my preference more and more.

[...]

PR #786, opened 2025-01-28
feat: add the "valid-license" rule

### LINKED ISSUE #830: 🚀 Feature: Bring in a valid-license rule
### Bug Report Checklist

### Overview

Following #42 -> #51: we've resolved to create a standalone `valid-license` rule. It should enforce that, _if_ the field exists in the `package.json` file, it's set to a valid value. It should do nothing if the field doesn't exist.

### Additional Info

Per discussion in #51, this won't be enabled by default at first. For this issue, we should:

💖

xenobytezero:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #830

## Overview

Adds a new rule that enforces the 'licence' value in the package.json to be a specific value, or one of a selection of values.

Proposing this as a new rule to eliminate my usage of [npm-package-json-lint](https://www.npmjs.com/package/npm-package-json-lint), since this was the only feature I needed from it.

I did have this as a custom rule in my own ESLint plugin, but I found out it's very difficult to [apply multiple plugins to the same file](https://github.com/eslint/eslint/discussions/18808) and I am already using this plugin.

This rule could expand in scope to allow for any value to be validated in the same way.


src/plugin.ts
JoshuaKGoldberg, line 1: Lots of unexpected changes in this file - could you please revert them?

https://www.joshuakgoldberg.com/blog/split-out-unrelated-changes

src/tests/rules/requires-license.test.ts
```diff
@@ -0,0 +1,61 @@
...
+                    allowed: convertAllowedValuesToString(expectedAllowedArr),
+                    multiple: multipleDescriptor(expectedAllowedArr),
+                },
+                messageId: expectedMessageId,
+            },
+        ],
+    };
+}
+
+function createValidCase(
```
JoshuaKGoldberg, line 38: -1 from me on case factories / functions like this. They do improve test readability when just thinking about a single file, but when lots of files have them, it gets hard to mentally swap between them. These abstractions also make the code harder to statically analyze.

Change request: inline the cases like the other files please

JoshuaKGoldberg (changes requested):
Thanks for sending this! This was a good prod to close out the conversation in #51 about how we want to approach all these requirements. I posted https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/51#issuecomment-2619890532.

Left a couple of initial review comments here in case you feel motivated to work on this before we're certain this is the direction for the package. Cheers!

michaelfaith (commented):
It looks like the rule and rule test files aren't named consistently with the rule itself `requires-license` vs `require-license`

src/tests/rules/requires-license.test.ts
xenobytezero: Changed. This was a holdover from my custom plugin, inlined everything to match the other tests

src/plugin.ts
xenobytezero: Looks like there was some confusion around tabs vs spaces between Prettier and editorconfig (they seem to specify different settings) so all the tabs in plugin.ts got nuked. Fixed.

xenobytezero:
> It looks like the rule and rule test files aren't named consistently with the rule itself `requires-license` vs `require-license`

Rookie error, fixed.

michaelfaith:
Thanks for you patience as we worked out the long-term strategy for the rules in this plugin. What we landed on in #52 is that we want to have separate rules for requiring a property and for validating the correctness of that property (if it exists).  That way people can use one or the other, or both. 

So for `require-licence`, the rule should just purely test for the existence of the property. And a separate rule `valid-licence` would check that if the property exists that it is valid. The current state of this PR has both requiredness and validation checks. So, in order for this to land, it would need to be reduced to just the check for requiredness. Does that make sense?

michaelfaith:
Actually, on second though, why don't we convert this one into the `valid-license` rule, since most of the logic is oriented towards that?  The main difference now, though, is that there's no check for required-ness, since the other rule will handle that.  Let me know if any of that is unclear and I'll help clear things up.

xenobytezero:
> Actually, on second though, why don't we convert this one into the `valid-license` rule, since most of the logic is oriented towards that? The main difference now, though, is that there's no check for required-ness, since the other rule will handle that. Let me know if any of that is unclear and I'll help clear things up.

No problem, can rework this rule to check for validity. 

- Obviously checking validity would require existence of the property in order to validate the value, and from what I can gather in reading #51 the rule should also fail for non-existance? 
- Should it fail in the way it does now? 
- Should I split out the `require` logic into it's own thing since most of the `requires` rules will be very similar?

michaelfaith:
> and from what I can gather in reading [Bring in a "require-*"/"no-*" kind of rule #51](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/51) the rule should also fail for non-existance?

Actually that went back and forth a couple of times (my bad 😅).  Where we landed is that `valid-` rules will _not_ fail if the property isn't present.  That will be the responsibility of the `require-` rules.  `valid-` rules will just validate that, if the property is present, its value is valid. It won't do anything if there isn't a property with that name.

xenobytezero:
Just working on this now, looks like the changes would be essentially cosmetic to match what was discussed.

The rule currently has a check that if the property is present that it also must be a string. I don't think there is anything in the current discussion about where type checking of the value should live in `require-` or `valid-`?

michaelfaith:
The require- rules won't do any checking of the value. It'll just simply check whether the property is presentor not.

src/rules/valid-license.ts
```diff
@@ -0,0 +1,80 @@
...
+			category: "Best Practices",
+			description:
+				"Enforce the 'license' field to be a specific value/values",
+			recommended: false,
+		},
+		messages: {
+			invalidValue: '"license" must be{{multiple}} {{allowed}}"',
+			nonString: '"license" must be a string"',
+		},
+		schema: {
```
JoshuaKGoldberg, line 60: [Feature] If this rule is to be enabled in the recommended config -which would be great!- then we're going to need it to not be prescriptive on which license(s) are allowed. Any "valid" license should be allowed unless the user has customized the rule. "Valid" here could mean one of several things, depending on how strict the user wants to be. In increasing order of strictness...

1. 🍏 Any string value
2. 🍌 Any [SPDX license ID](https://spdx.org/licenses) (on npm: [`spdx-license-ids`](https://www.npmjs.com/package/spdx-license-ids)) or [SPDX license expression syntax](https://spdx.dev/use/specifications) referring to SPDX license IDs
3. 🌸 The same as 2., but only OSI approved license IDs
4. 🐉 The same as 2., but only FSF Free/Libre approved license IDs
5. 🥚 >=1 specific license IDs (what the PR has right now)

I think any of 2 🍌, 3 🌸, 4 🐉, and 5 🥚 would be fine as followup issues (with the caveat that 3 🌸 and 4 🐉 depend on 2 🍌). Most users would be satisfied with just 1 🍏 checking that it's a string.

But, the options right now don't allow 1 🍏. They require 5 🥚, passing in a specific list of IDs. Could you please make the 5 🥚 specific license ID checking optional? As in, if the user doesn't provide any licenses, the rule should only check 1 🍏 that it's a valid string.

Reference: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#license

src/tests/rules/valid-license.test.ts
```diff
@@ -0,0 +1,74 @@
+import { generateReportData, rule } from "../../rules/valid-license.js";
+import { ruleTester } from "./ruleTester.js";
+
+ruleTester.run("valid-license", rule, {
+    invalid: [
+        // Invalid with single valid value
```
JoshuaKGoldberg, line 6: [Docs] Nit: IMO these comments aren't worth keeping up-to-date. ESLint rule tests tend to get very nuanced and tricky over time. I've seen suits with many dozens or even hundreds of tests that exercise various edge cases on themes. Personally I'd remove this:

```suggestion
```

If you _really_ want to include a name per test, the proper way would be with a `name` property per https://eslint.org/docs/latest/integrate/nodejs-api#ruletester. 

Not a blocker, just a preference. 🙂

src/plugin.ts
JoshuaKGoldberg, line 1: [Formatting] Still some editorconfig issues here?

JoshuaKGoldberg (changes requested):
Progress 🚀

JoshuaKGoldberg:
👋 Just checking in @xenobytezero, is this still something you have time for?

src/rules/valid-license.ts
```diff
@@ -0,0 +1,88 @@
...
+							},
+							type: "array",
+						},
+						{
+							type: "string",
+						},
+					],
+				},
+			],
+			type: "array",
```
JoshuaKGoldberg: [Feature] Most other rules in this plugin use an object for properties, like `order-properties`:

https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/5b5c2989dda36b43e85a14056120e863ef4fec3f/docs/rules/order-properties.md#L42-L52

Inline array/string options are nice and succinct but don't scale well when other options get added in. Especially for rules like this one that have nuanced options suggested (https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/786#discussion_r1957414117).

cc @michaelfaith - I can't find us establishing a standard of using options properties. WDYT? 

Left to my own devices I would request this to be an options object the way e.g. `order-properties` does? Maybe, `allow` for the option name?

JoshuaKGoldberg (changes requested):
LGTM! Sorry for not noting the options shape before, I'd missed it 😕.

I also took the liberty of merging in from `main`, running `pnpm format --write`, and fixing a region-specific spell checker complaint. The last CI complaint should be resolved if you add an `### Options` to `order-properties.md`.

cc @michaelfaith - this might be an instance of cspell getting in the way 😄

src/rules/valid-license.ts
michaelfaith: > I can't find us establishing a standard of using options properties. WDYT?

I don't believe we've formalized anything about this, but I like the recommendation.  Less likely to have to introduce breaking changes down the road if we want to expand the rule's functionality.

src/rules/valid-license.ts
```diff
@@ -0,0 +1,88 @@
...
+	};
+}
+
+function forceToArray<T>(item: T | T[]): T[] {
+	return Array.isArray(item) ? item : [item];
+}
+
+export const rule = createRule<Options>({
+	create(context) {
+		const ruleOptions = context.options[0];
```
michaelfaith: Nit: this intermediate variable doesn't seem to be adding much value.

src/rules/valid-license.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { createRule } from "../createRule.js";
+
+export type AllowedValues = string | string[] | undefined;
+export type Options = [AllowedValues];
+
+export function generateReportData(allowed: string[]) {
+	return {
+		allowed: allowed.map((v) => `'${v}'`).join(","),
+		multiple: allowed.length > 1 ? "one of" : "",
```
michaelfaith: I believe this needs to be `" one of"` (with a preceding space) to avoid conjoining with "be" in the message.

michaelfaith (changes requested):
One minor nit and one bug from my side.  Thanks for continuing to work on this.

michaelfaith:
Closing in favor of https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1174, which uses `validateLicense` from `package-json-validator` to validate the license string against known SPDX licenses.

One thing about the implementation in this PR @xenobytezero, that I think might be worth carrying over to the linked version of the rule, is the ability to further constrain the licenses to some subset of valid licenses defined in an option.  If you'd be up for adding that as an option, to the new rule once #1174 merges, we'd certainly welcome a new issue for that.

PR #851, opened 2025-02-02
feat(require-author): add new `require-author` rule

### LINKED ISSUE #795: 🚀 Feature: Bring in a require-author rule
### Bug Report Checklist

### Overview

Following #42 -> #51: we've resolved to create a standalone `require-author` rule. It should enforce that the field exists in the `package.json` file.

### Additional Info

💖

michaelfaith:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #795 

## Overview

This change adds the first or our new `require-` rules, and creates some foundational plumbing to make it super easy to add new require rules.  I've only done author in this PR.  Assuming this all looks good, we can quickly knock all the rest of the require rules in one shot.

### DIFF
--- a/src/rules/require-properties.ts
+++ b/src/rules/require-properties.ts
@@ -2,25 +2,21 @@ import type { PackageJsonRuleModule } from "../createRule.js";
 
 import { createRequirePropertyRule } from "../utils/createRequirePropertyRule.js";
 
-interface PropertyRule {
-	isRecommended: boolean;
-	propertyName: string;
-}
-
-// List of all properties we want to create require- rules for.
+// List of all properties we want to create require- rules for,
+// in the format [propertyName, isRecommended]
 const properties = [
-	{ isRecommended: false, propertyName: "author" },
-] satisfies PropertyRule[];
+	["author", false],
+	// TODO: More to come!
+] satisfies [string, boolean][];
 
-/** All require- flavor of rules */
-const rules: Record<string, PackageJsonRuleModule> = {};
-
-// Create all require- rules
-for (const { isRecommended, propertyName } of properties) {
-	rules[`require-${propertyName}`] = createRequirePropertyRule(
-		propertyName,
-		isRecommended,
-	);
-}
-
-export { rules };
+/** All require- flavor rules */
+export const rules = properties.reduce<Record<string, PackageJsonRuleModule>>(
+	(acc, [propertyName, isRecommended]) => {
+		acc[`require-${propertyName}`] = createRequirePropertyRule(
+			propertyName,
+			isRecommended,
+		);
+		return acc;
+	},
+	{},
+);
--- a/src/rules/require-properties.ts
+++ b/src/rules/require-properties.ts
@@ -1,0 +1,26 @@
+import type { PackageJsonRuleModule } from "../createRule.js";
+
+import { createRequirePropertyRule } from "../utils/createRequirePropertyRule.js";
+
+interface PropertyRule {
+	isRecommended: boolean;
+	propertyName: string;
+}
+
+// List of all properties we want to create require- rules for.
+const properties = [
+	{ isRecommended: false, propertyName: "author" },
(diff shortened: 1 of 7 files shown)


src/rules/require-properties.ts
JoshuaKGoldberg, line 1: [Style] A couple of quick thoughts here:

* I don't think the `//`/`/*` comments are worth their weight; IMO their content is inferable from the file
* We could trim this down just a bit by using an `as const` and an array of tuples

WDYT of this?

```ts
const properties = [
	["author", false],
	// TODO: more to come :)
] as const;

export const rules = Object.fromEntries(
	properties.map(([propertyName, recommended]) => [
		`require-${propertyName}`,
		createRequirePropertyRule(propertyName, recommended),
	]),
);
```

src/utils/createRequirePropertyRule.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Given a top-level property name, create a rule that requires that property to be present.
+ * Optionally, include it in the recommended config.
+ */
+export const createRequirePropertyRule = (
+	propertyName: string,
+	isRecommended = false,
+) => {
+	return createRule({
+		create(context) {
+			let hasSeen = false;
```
JoshuaKGoldberg, line 13: [Refactor] This _"wait until the end"_ strategy adds a bit more complexity than we need. Since we know the property must be a child of the object expression, we can directly check its properties:

```ts
"Program > JSONExpressionStatement > JSONObjectExpression"(
	node: JsonAST.JSONObjectExpression,
) {
	if (
		!node.properties.some(
			(property) =>
				isJSONStringLiteral(property.key) &&
				property.key.value === propertyName,
		)
	) {
		context.report({
			data: { property: propertyName },
			messageId: "missing",
			node: context.sourceCode.ast,
		});
	}
},
```

src/tests/rules/require-author.test.ts
```diff
@@ -0,0 +1,35 @@
...
+					data: { property: "author" },
+					line: 1,
+					messageId: "missing",
+				},
+			],
+		},
+	],
+	valid: [
+		`{ "main": "./index.js", "author": "Sophie Trudeau" }`,
+		`{ "author": "Jessica Moss" }`,
```
JoshuaKGoldberg, line 33: [Question] What is the reference with these names? Google isn't helping me 😄

src/tests/rules/require-author.test.ts
JoshuaKGoldberg, line 1: [Testing] Thinking on odd cases that can happen in the AST that should be tested:

* In the types, `JSONProperty`'s key is `JSONIdentifier | JSONStringLiteral | JSONNumberLiteral`
* Nested object with and without `author`, to make sure only the root-level one is checked

JoshuaKGoldberg (changes requested):
👍 from me on this approach, agreed it'll be handy as we scale to more properties. Nice!

Requesting changes on simplifying the two main files a bit and some more testing.

src/tests/rules/require-author.test.ts
michaelfaith: They're both members of an amazing post-rock band, Thee Silver Mt. Zion,  which spun out of pioneer post-rock band Godspeed You! Black Emperor.  Sophie's also a member of GYBE, and Jessica is married to Efrim who's also in GYBE.  I originally had two other people, but cspell didn't recognize the words "Reznor", "Efrim", or "Menuck", so I went with these 😊

src/rules/require-properties.ts
michaelfaith: I like the idea of using tuples here.  I changed the `properties` object over to that, but kept `satisfies` in there to enforce a type expectation.  I also went with a reducer, since that felt a bit more readable / communicative.

JoshuaKGoldberg: Also surfacing from a 1:1 conversation: I'm not against trying out more comments 😄

PR #876, opened 2025-02-05
chore: add dev script

chouchouji:
## PR Checklist

## Overview

I added two rules for this repo. My work process is to first complete the code, and then use this project for debugging. I will frequently change the code. But I don't want to  run `pnpm build` after updating code every time. So adding a `dev` script is helpful for me to work better.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
 	],
 	"scripts": {
 		"build": "tsup",
+		"dev": "tsup --watch",
 		"format": "prettier \"**/*\" --ignore-unknown",
 		"lint": "eslint . --max-warnings 0",
 		"lint:eslint-docs": "pnpm update:eslint-docs --check",


package.json
```diff
@@ -37,6 +37,7 @@
 	],
 	"scripts": {
 		"build": "tsup",
+		"dev": "tsup --watch",
```
JoshuaKGoldberg, line 40: Ehh, I'm a little -1 on this:

* I don't like adding extra properties to config files if they're not needed
* There are other `tsup` properties to consider: e.g. sometimes I add `--no-dts`

🤷 I'm not hard-set on vetoing this but not really seeing the value of this over running `pnpm build --watch`?

michaelfaith: I have to agree. I will often run tests in single run mode by passing `run` to the existing test command. Rather than have a whole new script for it. @chouchouji what value do you see a new script adding that passing `--watch` to the existing command doesn't provide?

chouchouji: I think `pnpm build --watch` is good for me.

PR #903, opened 2025-02-19
feat(valid-scripts): add valid-scripts rule (#839)

### LINKED ISSUE #839: 🚀 Feature: Bring in a valid-scripts rule
### Bug Report Checklist

### Overview

Following #42 -> #51: we've resolved to create a standalone `valid-scripts` rule. It should enforce that, _if_ the field exists in the `package.json` file, it's set to a valid value. It should do nothing if the field doesn't exist.

### Additional Info

Per discussion in #51, this won't be enabled by default at first. The `valid-package-def` rule already contains this logic using [`package-json-validator`](https://github.com/JoshuaKGoldberg/package-json-validator). For this issue, we should:

1. File an issue on https://github.com/JoshuaKGoldberg/package-json-validator to create a standalone API to validate this specific field
2. Send a PR to add API to `package-json-validator`
3. Once a version of `package-json-validator` is released with this field's API, mark this issue as unblocked
4. Send a PR to this repo adding in a standalone rule for this field - without adding it to the recommended config

💖

ic4l4s9c:
This change adds a new `valid-scripts` rule, which is also included in the `recommended` config.

## PR Checklist

-   [x] Addresses an existing open issue: fixes #839 

## Overview

:black_heart:

### DIFF
--- a/docs/rules/valid-scripts.md
+++ b/docs/rules/valid-scripts.md
@@ -1,0 +1,54 @@
+# valid-scripts
+
+💼 This rule is enabled in the ✅ `recommended` config.
+
+<!-- end auto-generated rule header -->
+
+This rule applies two validations to each property in `"scripts"` field:
+
+-   It must be a string rather than any other data type
+-   It should not be empty
+
+Example of **incorrect** code for this rule:
+
+```json
+{
+	"scripts": null
+}
+```
+
+```json
+{
+	"scripts": {}
+}
+```
+
+```json
+{
+	"scripts": {
+		"test": ""
+	}
+}
+```
+
+Example of **correct** code for this rule:
+
+```json
+{
+	"scripts": {
+		"build": "tsup",
+		"format": "prettier \"**/*\" --ignore-unknown",
+		"lint": "eslint . --max-warnings 0",
+		"lint:eslint-docs": "pnpm update:eslint-docs --check",
+		"lint:knip": "knip",
+		"lint:md": "markdownlint \"**/*.md\" \".github/**/*.md\"",
+		"lint:packages": "pnpm dedupe --check",
+		"lint:spelling": "cspell \"**\" \".github/**/*\"",
+		"prepare": "husky",
+		"should-semantic-release": "should-semantic-release --verbose",
+		"test": "vitest",
+		"tsc": "tsc",
+		"update:eslint-docs": "eslint-doc-generator --rule-doc-title-format name"
+	}
+}
+```
--- a/src/rules/valid-scripts.ts
+++ b/src/rules/valid-scripts.ts
@@ -1,0 +1,69 @@
+import type { AST as JsonAST } from "jsonc-eslint-parser";
+
+import * as ESTree from "estree";
+
+import { createRule } from "../createRule.js";
+
+export const rule = createRule({
+	create(context) {
+		return {
+			"Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.value=scripts]"(
+				node: JsonAST.JSONProperty,
+			) {
+				const { value } = node;
+				if (value.type !== "JSONObjectExpression") {
+					context.report({
+						messageId: "incorrectTypeScriptsField",
+						node: node.value as unknown as ESTree.Node,
+					});
+					return;
+				}
+				if (value.properties.length === 0) {
+					context.report({
+						messageId: "emptyScriptsField",
+						node: node.value as unknown as ESTree.Node,
+					});
+					return;
+				}
+				for (const prop of value.properties) {
+					const { value } = prop; // key,
+					if (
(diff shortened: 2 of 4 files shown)


src/rules/valid-scripts.ts
```diff
@@ -0,0 +1,69 @@
...
+		docs: {
+			category: "Best Practices",
+			description: "Enforce that package scripts are valid commands",
+			recommended: true,
+		},
+		messages: {
+			emptyScript:
+				'values contained in "scripts" object should not be empty',
+			emptyScriptsField: '"scripts" field should not be empty.',
+			incorrectTypeScript:
```
JoshuaKGoldberg, line 61: [Praise] Haha, "TypeScript". I thought this was cute. 🙂

src/rules/valid-scripts.ts
```diff
@@ -0,0 +1,69 @@
...
+					const { value } = prop; // key,
+					if (
+						value.type !== "JSONLiteral" ||
+						typeof value.value !== "string"
+					) {
+						context.report({
+							messageId: "incorrectTypeScript",
+							node: node.value as unknown as ESTree.Node,
+						});
+						return;
```
JoshuaKGoldberg, line 38: [Question] Why `return;`? If multiple scripts are incorrect, we'd want to get reports on all of them I think.

E.g. here I'd expect the values of both `script1` and `script3` to have complaints:

```json
"scripts": {
    "script1": "",
    "script2": "...",
    "script3": "",
}
```

(here and the other `return`)

src/tests/rules/valid-scripts.test.ts
```diff
@@ -0,0 +1,61 @@
...
+			errors: [{ messageId: "emptyScript" }],
+		},
+		{
+			code: `{ "scripts": { "test": null }}`,
+			errors: [{ messageId: "incorrectTypeScript" }],
+		},
+		{
+			code: `{ "scripts": { "test": undefined }}`,
+			errors: [{ messageId: "incorrectTypeScript" }],
+		},
```
JoshuaKGoldberg, line 43: [Testing] No need, `undefined` isn't valid JSON.

```suggestion
```

I'm actually surprised this doesn't cause any _"hey you have invalid stuff"_ complaint in the tester...

src/tests/rules/valid-scripts.test.ts
```diff
@@ -0,0 +1,61 @@
...
+		{
+			code: `{ "scripts": { "test": {} }}`,
+			errors: [{ messageId: "incorrectTypeScript" }],
+		},
+		{
+			code: `{ "scripts": { "test": 678 }}`,
+			errors: [{ messageId: "incorrectTypeScript" }],
+		},
+	],
+	valid: [
```
JoshuaKGoldberg, line 53: [Testing] Some more valid cases to think of:

* No `"scripts"` key at all
* Empty `"scripts"`
* Just one value in `"scripts"`

src/tests/rules/valid-scripts.test.ts
```diff
@@ -0,0 +1,61 @@
+import { rule } from "../../rules/valid-scripts.js";
+import { ruleTester } from "./ruleTester.js";
+
+ruleTester.run("valid-scripts", rule, {
+	invalid: [
```
JoshuaKGoldberg, line 5: [Testing] Some more invalid cases to think of:

* 1 incorrect value + 2 correct values
* 2 incorrect values
* duplicate values
* 2 incorrect values + 1 correct value

src/rules/valid-scripts.ts
JoshuaKGoldberg, line 1: [Feature] Fun fact, it is technically valid JSON to have duplicate keys:

```json
"scripts": {
  "this-is": "fine",
  "this-is": "fine"
}
```

cc @michaelfaith to check me here, but IMO this should be reported on by `valid-scripts`.

docs/rules/valid-scripts.md
```diff
@@ -0,0 +1,54 @@
+# valid-scripts
+
+💼 This rule is enabled in the ✅ `recommended` config.
+
+<!-- end auto-generated rule header -->
+
+This rule applies two validations to each property in `"scripts"` field:
```
JoshuaKGoldberg, line 7: [Docs] Just nitpicking a little bit for clarity, WDYT?

```suggestion
This rule checks that if a `"scripts"` property exists, it is a non-empty object.
Additionally, for each of its script properties:
```

JoshuaKGoldberg (changes requested):
😱 sorry for missing this for multiple weeks! This generally looks great to me - mostly small touchups. I think the only big point is whether we should validate that multiple keys aren't identical (IMO we should, but open to thoughts?).

JoshuaKGoldberg:
👋 Just checking in @ic4l4s9c, is this still something you have time for?

PR #998, opened 2025-04-05
feat: new `restrict-dependency-ranges` rule

### LINKED ISSUE #959: 🚀 Feature: add restrict dependency ranges rule
### Bug Report Checklist

### Overview

This is a subset of the feature request in #54, specifically for introducing a rule to allow for restricting dependency ranges, similar to [restrict-ranges](https://github.com/kellyselden/eslint-plugin-json-files/blob/main/docs/rules/restrict-ranges.md) from `eslint-plugin-json-files`.

Proposing a rule named `restrict-dependency-ranges` with the following options (or an array of these options):

- `forDependencyTypes: ('dependencies' |  'devDependencies' | 'optionalDependences' | 'peerDependencies')[] ` - Apply these options to a specific type of dependencies
- `forPackages: string[]` - Specific packages or regex patterns to apply these options to.
- `forVersions: string` - Semver range to apply these options to (e.g. `<1`)
- `rangeType: 'caret' | 'tilde' | 'pin' | ('caret' | 'tilde' | 'pin)[]` (**required**) - Specifies the type of range you want to enforce (^, ~, or no range) on the matched dependencies

michaelfaith:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #959

## Overview

This change adds a new rule that allows you to restrict usage of semver ranges on dependencies using any of the following matching options:
- by dependency name or regex pattern
- by version
- by dependency type

Closes #959

### DIFF
--- a/docs/rules/restrict-dependency-ranges.md
+++ b/docs/rules/restrict-dependency-ranges.md
@@ -1,0 +1,114 @@
+# restrict-dependency-ranges
+
+💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
+
+This rule allows you to require that specific dependencies use a particular kind
+of semver range (e.g. `^`). There are several options for specifying which dependencies
+a range type restriction should be applied to, including dependency type,
+package name (or name regex pattern), and version range (e.g. '<1').
+
+If you provide multiple options and a dependency matches more than one of the
+options, the last option that matches will take precedent for that dependency.
+This allows you to define more general rules that apply to all dependencies (or large
+groups of dependencies), and then define more granular options to apply exceptions
+or focus on some subset of dependencies.
+
+## Options
+
+`type DependencyType = 'dependencies' | 'devDependencies' | 'optionalDependencies' | 'peerDependencies';`\
+`type RangeType = 'caret' | 'pin' | 'tilde';`
+
+| Name                 | Type                     | Required |
+| :------------------- | :----------------------- | :------- |
+| `forDependencyTypes` | DependencyType[]         |          |
+| `forPackages`        | String[]                 |          |
+| `forVersions`        | String                   |          |
+| `rangeType`          | RangeType \| RangeType[] | Yes      |
+
+You can provide a single options object consisting of the above, or an array
+of such objects.
+
+### `forDependencyTypes`
+
+You can use this to apply a range type restriction for an entire group of dependencies
+by which type of dependencies they belong to.
+
+Options are
+
+- dependencies
+- devDependencies
+- optionalDependencies
+- peerDependencies
+
+### `forPackages`
+
+This can be the exact name of a package, or a regex pattern used to match a
+group of packages by name (e.g. `@typescript-eslint/*`).
+
+### `forVersions`
+
+You can use this to apply a restriction to a specific semver range. For example,
+a common use case is to pin "unstable" dependencies (packages that have
+a version in the `0.x.x` range). You can do this by setting `forVersions` to `'<1'`.
+
+### `rangeType`
+
+This is the only required option, and identifies which range type or types you
+want to apply to packages that match any of the other match options (or all
+dependencies if no other options are provided).
+
+## Example
+
+```ts
+export default [
+	{
+		"package-json/restrict-dependency-ranges": [
+			"error",
+			[
+				// Apply base requirement that all deps should use ^
+				{
+					rangeType: "caret",
+				},
+
+				// Restrict typescript to tilde ranges
+				{
+					forPackages: "typescript",
+					rangeType: "tilde",
+				},
+
+				// Require packages that have 0.x.x versions are pinned
+				{
+					forVersions: "<1",
+					rangeType: "pin",
+				},
+			],
+		],
+	},
+];
+```
+
+Example of **incorrect** code for the above configuration:
+
+```json
+{
+	"devDependencies": {
+		"eslint": "^9.18.0",
+		"markdownlint": "^0.37.4",
+		"typescript": "^5.8.0"
+	}
+}
+```
+
+Example of **correct** code for the above configuration:
+
+```json
+{
+	"devDependencies": {
+		"eslint": "^9.18.0",
+		"markdownlint": "0.37.4",
+		"typescript": "~5.8.0"
+	}
+}
+```
--- a/docs/rules/restrict-dependency-ranges.md
+++ b/docs/rules/restrict-dependency-ranges.md
@@ -4,10 +4,8 @@
 
 <!-- end auto-generated rule header -->
 
-This rule allows you to require that specific dependencies use a particular kind
-of semver range (e.g. `^`). There are several options for specifying which dependencies
-a range type restriction should be applied to, including dependency type,
-package name (or name regex pattern), and version range (e.g. '<1').
+This rule allows you to require that specific dependencies use a particular kind of semver range (e.g. `^`).
+There are several options for specifying which dependencies a range type restriction should be applied to, including dependency type, package name (or name regex pattern), and version range (e.g. '<1').
 
 ```ts
 export default [
@@ -24,11 +22,8 @@ export default [
 ];
 ```
 
-If you provide multiple options and a dependency matches more than one of the
-options, the last option that matches will take precedent for that dependency.
-This allows you to define more general rules that apply to all dependencies (or large
-groups of dependencies), and then define more granular options to apply exceptions
-or focus on some subset of dependencies.
+If you provide multiple options and a dependency matches more than one of the options, the last option that matches will take precedent for that dependency.
+This allows you to define more general rules that apply to all dependencies (or large groups of dependencies), and then define more granular options to apply exceptions or focus on some subset of dependencies.
 
 ## Options
 
@@ -42,13 +37,11 @@ or focus on some subset of dependencies.
 | `forVersions`        | string                   |          |
 | `rangeType`          | RangeType \| RangeType[] | Yes      |
 
-You can provide a single options object consisting of the above, or an array
-of such objects.
+You can provide a single options object consisting of the above, or an array of such objects.
 
 ### `forDependencyTypes`
 
-You can use this to apply a range type restriction for an entire group of dependencies
-by which type of dependencies they belong to.
+You can use this to apply a range type restriction for an entire group of dependencies by which type of dependencies they belong to.
 
 Options are
 
@@ -74,8 +67,7 @@ export default [
 
 ### `forPackages`
 
-This can be the exact name of a package, or a regex pattern used to match a
-group of packages by name (e.g. `@typescript-eslint/*`).
+This can be the exact name of a package, or a regex pattern used to match a group of packages by name (e.g. `@typescript-eslint/*`).
 
 ```ts
 export default [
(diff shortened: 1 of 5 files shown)


docs/rules/restrict-dependency-ranges.md
JoshuaKGoldberg, line 1: [Docs] There's a lot of text in this file before the final examples. I find it hard to conceptualize what a rule does from just text descriptions and tables alone. It would also be nice to have examples diving into each of the options: e.g. how to use the regex form of `forPackages`.

Proposal: how about...

* Starting with a very straightforward example after the first paragraph
* Including dedicated examples within each of the options

docs/rules/restrict-dependency-ranges.md
```diff
@@ -0,0 +1,114 @@
...
+## Options
+
+`type DependencyType = 'dependencies' | 'devDependencies' | 'optionalDependencies' | 'peerDependencies';`\
+`type RangeType = 'caret' | 'pin' | 'tilde';`
+
+| Name                 | Type                     | Required |
+| :------------------- | :----------------------- | :------- |
+| `forDependencyTypes` | DependencyType[]         |          |
+| `forPackages`        | String[]                 |          |
+| `forVersions`        | String                   |          |
```
JoshuaKGoldberg, line 27: [Docs] To be more correct with [`@typescript-eslint/no-wrapper-object-types`](https://typescript-eslint.io/rules/no-wrapper-object-types):

```suggestion
| `forPackages`        | string[]                 |          |
| `forVersions`        | string                   |          |
```

docs/rules/restrict-dependency-ranges.md
```diff
@@ -0,0 +1,114 @@
...
+			"error",
+			[
+				// Apply base requirement that all deps should use ^
+				{
+					rangeType: "caret",
+				},
+
+				// Restrict typescript to tilde ranges
+				{
+					forPackages: "typescript",
```
JoshuaKGoldberg, line 77: [Docs]

```suggestion
					forPackages: ["typescript"],
```

src/rules/restrict-dependency-ranges.ts
```diff
@@ -0,0 +1,293 @@
...
+	forDependencyTypes?: string[];
+	forPackages?: string[];
+	forVersions?: string;
+	rangeType: RangeType | RangeType[];
+}
+
+type Options = [Option | Option[] | undefined];
+
+const RANGE_TYPES = ["caret", "pin", "tilde"];
+type RangeType = "caret" | "pin" | "tilde";
```
JoshuaKGoldberg, line 26: [Style] Nit, so the code doesn't have to repeat itself:

```suggestion
const RANGE_TYPES = ["caret", "pin", "tilde"] as const;
type RangeType = (typeof RANGE_TYPES)[number];
```

src/rules/restrict-dependency-ranges.ts
```diff
@@ -0,0 +1,293 @@
...
+		/^(workspace:)?(\^|~|<=?|>=?)?/,
+		`$1${SYMBOLS[rangeType]}`,
+	);
+};
+
+/**
+ * Check if the version is in a form that this rule supports.
+ */
+const isVersionSupported = (version: string): boolean => {
+	if (version === "*" || /^workspace:[*^~]$/.test(version)) {
```
JoshuaKGoldberg, line 98: > `version === "*"`

[Testing] No tests fail if this check is removed. Missing unit test(s)?

src/rules/restrict-dependency-ranges.ts
```diff
@@ -0,0 +1,293 @@
...
+		return true;
+	}
+
+	const rawVersion = version.replace(/^workspace:/, "");
+	return !!semver.valid(rawVersion) || !!semver.validRange(rawVersion);
```
JoshuaKGoldberg, line 103: > `!!semver.valid(rawVersion) ||`

[Testing] No tests fail if this check is removed. Missing unit test(s)?

src/rules/restrict-dependency-ranges.ts
```diff
@@ -0,0 +1,293 @@
...
+					key: JsonAST.JSONStringLiteral;
+					value: JsonAST.JSONObjectExpression;
+				},
+			) {
+				const dependencyType = node.key.value;
+
+				// If this isn't a group of dependencies, skip it entirely
+				if (!DEPENDENCY_TYPES.includes(dependencyType)) {
+					return;
+				}
```
JoshuaKGoldberg, line 136: [Testing] No tests fail if this check is removed. Missing unit test(s)?

src/rules/restrict-dependency-ranges.ts
```diff
@@ -0,0 +1,293 @@
...
+
+				// Loop through all dependencies in the group
+				for (const property of node.value.properties) {
+					// If either the key or value aren't strings, this isn't a valid dependency, so move on.
+					if (
+						!isJSONStringLiteral(property.key) ||
```
JoshuaKGoldberg, line 142: [Testing] No tests fail if this check is removed. Missing unit test(s)?

src/rules/restrict-dependency-ranges.ts
```diff
@@ -0,0 +1,293 @@
...
+							continue;
+						}
+						if (options.forPackages) {
+							let isMatch = false;
+							for (const packageNamePattern of options.forPackages) {
+								if (new RegExp(packageNamePattern).test(name)) {
+									isMatch = true;
+									break;
+								}
+							}
```
JoshuaKGoldberg, line 184: [Style] Nit: we can avoid the extra mutable variable with a little `.some`:

```suggestion
							const isMatch = options.forPackages.some(
								(packageNamePattern) =>
									new RegExp(packageNamePattern).test(name),
							);
```

src/rules/restrict-dependency-ranges.ts
```diff
@@ -0,0 +1,293 @@
...
+		if (!context.options[0]) {
+			return {};
+		}
+
+		const optionsArray = Array.isArray(context.options[0])
+			? [...context.options[0]]
+			: [context.options[0]];
+
+		// Reverse the array, so that subsequent options override previous ones
+		optionsArray.reverse();
```
JoshuaKGoldberg, line 122: [Refactor] Later on, the code calls `new RegExp(packageNamePattern)` for every `forPackages` entry x for every package. If a config has more complex patterns and a lot of package deps then that's going to end up being a lot of regexp creation and `.test()`ing of those newly created regexes.

How about pre-processing options early to normalize them to the format we expect? This'll also let us avoid the `rangeTypes` normalization later on:

```suggestion
		// Reverse the array, so that subsequent options override previous ones
		const optionsProvided = Array.isArray(context.options[0])
			? [...context.options[0]].reverse()
			: [context.options[0]];

		const options = optionsProvided.map((option) => ({
			...option,
			forPackages: option.forPackages?.map(
				(pattern) => new RegExp(pattern),
			),
			rangeTypes: Array.isArray(option.rangeType)
				? option.rangeType
				: [option.rangeType],
		}));
```

JoshuaKGoldberg (changes requested):
🚀 The core logic seems very reasonable and good to me, really nice job! I think it makes sense to go in order like this and check the first (last, reversed) matching dependency.

Mainly requesting changes on:

* Docs: adding some more examples to make usage more clear
* Testing: some of the lines don't fail any tests for me when removed locally
* Some suggestions around refactors

WDYT?

docs/rules/restrict-dependency-ranges.md
michaelfaith: Updated

docs/rules/restrict-dependency-ranges.md
michaelfaith: Updated

docs/rules/restrict-dependency-ranges.md
michaelfaith: Updated

src/rules/restrict-dependency-ranges.ts
michaelfaith: Updated

michaelfaith:
From the "Lint Markdown" step: `Each sentence should be on its own line`  
Really?  That seems super heavy-handed, with no perceived value...

JoshuaKGoldberg:
😄 I do feel strongly in favor about it, I think it's really nice to have a 1:1 of "line that can be diffed in Git" to sentences. But also if you end up hating it enough to send a PR that removes it from this repo, I'll approve it. 

ACK that it's kind of inconvenient to not have it be integrated into ESLint; I'm hoping that later this year as `@eslint/markdown` gets fleshed out I can switch from Markdownlint to ESLint altogether...

michaelfaith:
> ACK that it's kind of inconvenient to not have it be integrated into ESLint

That actually touches on the second reason i'm not a fan of the rule haha. 

I have two main issues with it. For the first, i think we can generally agree on the following three statements?

1. Documentation is extremely valuable, but also not something that people always think to do (unless you maintain a project or are long in the tooth like me, it's often an afterthought). So  it's something we should generally incentivise contributors to do, by making it as easy as possible to create high quality documentation.
2. Writing good documentation is an art and not a science. Explaining something in a way that's easy for others to consume, understand, and relate to requires a fair amount of empathy and creative thinking. 
3. Writing prose one sentence per line does not come natural, and would not be how anyone would naturally write documentation unless they happen to know (and remember) this rule was in place. 

Those three things tell a story of us believing that something is truly valuable, but then creating friction around contributing it. More often than not, people contributing new code will hit up against that, and get frustrated (like i did 😆). That, then discourages people from even bothering. And all of that _could_ be ok, if we believe the benefits outweigh the cost. And i just don't see that being the case. If i think about the kinds of changes people might make to existing documentation, i don't really see what value having them one sentence per line brings in terms of diff review. Certainly not enough to outweigh the cost. For something like the dangling comma formatting rule (which promises a similar value proposition), that calculus is much more in favor of the rule, since it's something  we interact with way more.

And that brings me to the second point of grief: like dangling commas, this is a formatting concern. And i think we generally agree that formatting is the domain of prettier, not lint. This isn't something that feels like it should belong in any lint tool, whether it be markdownlint or eslint. In my mind, if it should exist at all, it should go in prettier.

Happy to be convinced otherwise, if i'm missing something, or not looking at this correctly. 😊

src/rules/restrict-dependency-ranges.ts
michaelfaith: I'm seeing 12 test failures when running without that condition

screenshot with failed test log

michaelfaith: Removed.  `semver.validRange` catches this.

src/rules/restrict-dependency-ranges.ts
michaelfaith: Removed.  `semver.validRange` catches this.

src/rules/restrict-dependency-ranges.ts
michaelfaith: Updated the "no dependencies" test

src/rules/restrict-dependency-ranges.ts
michaelfaith: Updated the "malformed dependencies" test

src/rules/restrict-dependency-ranges.ts
michaelfaith: Updated

src/rules/restrict-dependency-ranges.ts
michaelfaith: Updated

JoshuaKGoldberg:
You make a lot of good points! Agreed, getting folks to write docs in the first place >>> nitpicking the formatting of those docs 😄. Filed https://github.com/JoshuaKGoldberg/create-typescript-app/issues/2216.

JoshuaKGoldberg (approved):
💯 solid!

PR #1038, opened 2025-05-06
feat: update `unique-dependencies` rule to prevent duplicates in `dependencies` and `devDependencies`

### LINKED ISSUE #1007: 🐛 Bug: `unique-dependencies` doesn't flag duplicates between dependencies and devDependencies
### Bug Report Checklist

### Expected

I am specifying the same dependency in both `dependencies` and `devDependencies`, ie:
```json
{
  "name": "my-project",
  "dependencies": {
    "typescript": "^5.7.3",
  },
  "devDependencies": {
    "typescript": "^5.7.3",
  },
}
```

I would expect this to throw an error as we are using the same dependency in `dependencies` and `devDependencies`, where the dependency specified in `devDependencies` would be redundant.

### Actual

However, above `package.json` does not throw any error related to `package-json/unique-dependencies`. 

### Additional Info

I have tried to do some quick testing around, and managed to fix it quickly on my side, after slight modification on `src/rules/unique-dependencies.ts`.

Current code:
```ts
// Other code ...
  create(context) {
    function check(
      elements: (JsonAST.JSONNode | null)[],
      getNodeToRemove: (element: JsonAST.JSONNode) => JsonAST.JSONNode,
    ) {
      const seen = new Set();
        .filter(isNotNullish)
// Other code ...
```

Modified code (I moved the `const seen = new Set();` line out of the `check` function block):
```ts
// Other code ...
  create(context) {
    const seen = new Set();
    function check(
      elements: (JsonAST.JSONNode | null)[],
      getNodeToRemove: (element: JsonAST.JSONNode) => JsonAST.JSONNode,
    ) {
        .filter(isNotNullish)
// Other code ...
```

The problem seems to come from this line `const seen = new Set();`, which is being initialized inside the `check` function block, and as a result we failed to compare the list of dependencies included inside `dependencies` and `devDependencies`.

sunnytsang1998:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1007 

## Overview

Updated the `unique-dependencies` rule to check whether some dependencies are included multiple times unexpectedly. We have defined a list of dependency properties (currently set to only `dependencies` and `devDependencies`), where dependencies must be unique. If dependencies are being included multiple times here an error would be thrown.


JoshuaKGoldberg (changes requested):
This seems reasonable at first, but there aren't any unit tests verifying the behavior. Could you add some in please? I can review more deeply after. Thanks! ❤️

src/rules/unique-dependencies.ts
```diff
@@ -19,9 +19,14 @@ const dependencyPropertyNames = new Set([
 	"peerDependencies",
 ]);
 
+const requiredUniqueDependencies = new Set(["dependencies", "devDependencies"]);
```
michaelfaith, line 22: Similar to the other comment, this name doesn't really do much to explain its purpose.  Maybe something like `groupsRequiringGlobalUniqueness`?

src/rules/unique-dependencies.ts
```diff
@@ -19,9 +19,14 @@ const dependencyPropertyNames = new Set([
...
+
+// This keep tracks of `dependencies` or `devDependencies` seen
+const uniqueDependenciesSeen = new Set();
 export const rule = createRule({
 	create(context) {
 		function check(
+			property: string,
```
michaelfaith, line 29: Nit: recommend renaming this to `dependencyGroup`

michaelfaith (changes requested):
Thanks for taking this up!  In addition to adding tests, I have a few other minor requests. related to naming.

sunnytsang1998:
Thank you for the reviews! I have updated and added a test too.

src/tests/rules/unique-dependencies.test.ts
```diff
@@ -178,6 +178,38 @@ ruleTester.run("unique-dependencies", rule, {
...
+					suggestions: [
+						{
+							messageId: "remove",
+							output: `{
+		"dependencies": {
+			"abc": "1.2.3"
+		},
+		"devDependencies": {
+			
+		}
```
michaelfaith, line 212: This is a good addition.  Can we also include a `valid` test that has the same dependency in `devDependencies` and `peerDependencies` to verify we're not detecting a violation in that scenario?  Also, recommend adding another `invalid` case, similar to the above, but that also has `abc` in `peerDependencies`, so that we can validate that it's leaving that group alone.  What do you think?

src/rules/unique-dependencies.ts
```diff
@@ -31,7 +40,13 @@ export const rule = createRule({
 				.filter(isNotNullish)
 				.filter(isJSONStringLiteral)
 				.reverse()) {
-				if (seen.has(element.value)) {
+				if (groupsRequiringGlobalUniqueness.has(dependencyGroup)) {
+					if (globalUniqueDependenciesSeen.has(element.value)) {
+						report(element, elements);
```
michaelfaith, line 45: My main concern with this approach is that we're reporting in a somewhat naive way, simply based on order of the dependency groups.  When we were reporting _only_ intra-group dependencies then the order of dependency groups didn't matter, and the only thing that mattered was the order of the dependencies within a group.  That makes sense because you'd want to report and remove the earlier redundant instances, since the later instance would always be overriding it anyway (making the earlier instance the redundant one).  However, in this case, if `dependencies` are ordered before `devDependencies` then it's going to report the `devDependencies` instance.  If `devDependencies` are ordered before `dependencies` then it'll report the `dependencies` instance.  That's probably not what the user would want or expect.  It's unstable in that sense.  We should generally prefer the `dependencies` instance, since that's what will be installed by the consumers of the package.  So, the naive order-based approach doesn't really work very well for this use case.

src/tests/rules/unique-dependencies.test.ts
sunnytsang1998: Makes good sense to me. I have added tests to address this.

sunnytsang1998:
Added a new commit to address the above comments, and updated the logic for checking for unique global dependencies.

src/rules/unique-dependencies.ts
```diff
@@ -19,43 +19,86 @@ const dependencyPropertyNames = new Set([
...
 				}
+
+				// Next, check if we need to check for global unique dependencies.
+				if (
+					groupsRequiringGlobalUniqueness.some((groupsList) =>
+						groupsList.groups.includes(dependencyGroup),
+					)
+				) {
+					// We will always need to iterate through every item in this case, as we would want to report every duplication.
+					groupsRequiringGlobalUniqueness.forEach(
```
michaelfaith, line 65: Wouldn't we only want to do this on group lists that include the current dependencyGroup?  The if condition is saying "there's at least one group list in my array that includes the dependency group", but then here, you're looping through the entire array, rather than just the group list that includes the dependency group.  This also results in a traversal of the list twice, when it could just be one for loop.  Recommend refactoring this to be a single for loop, which has an if to check if that group list includes the dependency group.  Also, recommend a traditional for loop instead of using `.forEach()` since the latter is generally less performant.

src/rules/unique-dependencies.ts
```diff
@@ -19,43 +19,86 @@ const dependencyPropertyNames = new Set([
...
 	create(context) {
-		const globalUniqueDependenciesSeen = new Set();
+		// This is a list of groups that required uniqueness among all of them globally.
+		// Inside each object, groups defined the corresponding list of groups to check for.
+		// "remove" will determine in which dependency group should the duplicated dependency be removed.
+		// eg: if same dependency appears in "dependencies" and "deveDependencies" we keep that from "dependencies".
+		const groupsRequiringGlobalUniqueness = [
+			{
+				groups: ["dependencies", "devDependencies"],
+				remove: "devDependencies",
```
michaelfaith, line 31: The way this data structure works, a "group list" could only consist of two groups, right?  Any more than that, and you couldn't have a single string for `remove`.  Seems like a limitation to the api.  Not a deal breaker. Just calling it out that if we wanted to have global uniqueness across more than two groups, this logic would need to be refactored.

michaelfaith (changes requested):
The approach is mostly sound.  There's just one area where the logic doesn't quite connect, and a non-blocking observation about the data structure.

sunnytsang1998:
> 👋 Just checking in @sunnytsang1998, is this still something you have time for?

Hello, sorry for a slightly late reply! Have been busy recently and I afraid I don't really have time to work on this unfortunately.

PR #1048, opened 2025-05-17
 feat(sort-collections): support nested properties 

### LINKED ISSUE #1093: 🚀 Feature: [sort-collections] Support nested properties
### Feature Request Checklist

### Overview

Filing a backing issue for #1048: it'd be nice if the array option for [`sort-collections`](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/bae94b56e23f60e5f38c5e4b5260737c8740c81b/docs/rules/sort-collections.md) supported nested properties. As @andreww2012 noted:

> This can be useful for nested properties ordering, such as `pnpm.patchedDependencies`.

### Additional Info

🗂️

andreww2012:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1093

## Overview

Hello! This PR adds the ability for `sort-collections` rule to accept a nested property using a dot notation. This can be useful for nested properties ordering, such as `pnpm.patchedDependencies`.


src/tests/rules/sort-collections.test.ts
```diff
@@ -212,6 +239,16 @@ ruleTester.run("sort-collections", rule, {
...
+}`,
+		},
+		{
+			code: `{
+	"nested": {
+		"devDependencies": {
+			"typescript": "4.8.4",
+			"eslint-plugin-package-json": "0.31.0"
+		}
+	}
```
andreww2012, line 251: This test ensures that nested properties having the same names as root level ones are *no longer* touched.

src/rules/sort-collections.ts
```diff
@@ -1,125 +1,148 @@
...
-	"dependencies",
-	"peerDependencies",
-	"overrides",
+const defaultCollections = new Set([
 	"config",
+	"dependencies",
+	"devDependencies",
 	"exports",
-];
+	"overrides",
```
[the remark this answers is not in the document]
michaelfaith: Seems like a good idea

src/rules/sort-collections.ts
```diff
@@ -1,125 +1,148 @@
...
+						keyPartsReversed.push(currNode.key.value);
 					}
 				}
+				const key = keyPartsReversed.reverse().join(".");
+				if (!toSort.has(key)) {
+					return;
+				}
+
+				const currentOrder = collection.properties;
+				const scripts = new Set(
```
michaelfaith, line 57: Nit: I think is probably a holdover from when this was just sorting scripts.  Can we rename this to something more in line with its current purpose?  `properties` maybe?

michaelfaith (approved):
This looks great.  Nice work, and thanks for contributing.  Just a minor nit from me, though, not blocking.  
I do also agree with you that adding `pnpm.overrides` to the default collections would make sense and probably `resolutions` to cover `yarn` as well.  If you do add `pnpm.overrides` it would be good to add another `valid` test, to show that correctly ordered nested props that should be ordered, are treated as valid.

michaelfaith:
@JoshuaKGoldberg It looks like the compliance portion of the build is still failing when I "allows" the workflow to run.  It says `resource not accessible by integration`.  (it also incorrectly flagged the pr title, though that could be because it's not able to get the information).
 
screenshot of failed build

andreww2012:
Thank you for the review! I've decided to add quite a few new collections to sort by default, all of which expect keys being package names.

Added a couple of new tests as per your request and to cover the case of sorting a collection of the 3rd level.

docs/rules/sort-collections.md
```diff
@@ -98,13 +98,22 @@ Defaults:
...
+	"dependenciesMeta",
+	"devDependencies",
+	"exports",
+	"optionalDependencies",
 	"overrides",
-	"config",
-	"exports"
+	"peerDependencies",
+	"peerDependenciesMeta",
+	"pnpm.allowedDeprecatedVersions",
```
michaelfaith, line 110: Hmm, So, I looked these up on the `pnpm` docs, since I wasn't familiar with some of them, and discovered that all of these (including `overrides`) have been moved to the yaml settings file instead of the `package.json`.  So, it's making me question whether we should include these after all?  I don't know, I could go either way.  But since this feature was mostly motivated to support this pattern, perhaps it's still valuable?  What do you think?

andreww2012: pnpm of versions <=9, which support the fields I've added, is [>3 times as much](https://github.com/search?q=%22packageManager%22%3A+%22pnpm%409+path%3A%2F%5E.*package.json%24%2F+&type=code) as [pnpm 10](https://github.com/search?q=%22packageManager%22%3A+%22pnpm%4010+path%3A%2F%5E.*package.json%24%2F+&type=code) so I think it's not unreasonable to include them in the default list until old versions die. I'd be fine with either final decision, so perhaps a third opinion is needed 😄

src/rules/sort-collections.ts
```diff
@@ -1,125 +1,157 @@
...
+				) {
+					return;
+				}
+
+				const keyPartsReversed = [nodeKey.value];
+				for (
+					let currNode: JsonAST.JSONNode | null | undefined =
+						node.parent;
+					// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+					currNode;
```
JoshuaKGoldberg, line 50: hmm?

```suggestion
					// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
					currNode;
```

docs/rules/sort-collections.md
JoshuaKGoldberg: Yeah I think it'd be better to leave pnpm-specific additions out of this PR. pnpm is one of many userland "extensions" to the package.json format. Supporting every collection from every popular extension would be a lot of defaults to maintain.

Proposal: let's treat that as a separate change to be discussed in a separate issue?

src/tests/rules/sort-collections.test.ts
JoshuaKGoldberg, line 1: [Testing] Could you add tests for a user explicitly providing option(s) with >= `"."`? Test coverage for the new `.` nesting is only covering default options right now, but we'll want to make sure userland ones are working too.

src/rules/sort-collections.ts
```diff
@@ -1,125 +1,157 @@
...
-							return aKey > bKey ? 1 : -1;
-						}
-					});
 					if (
-						currentOrder.some(
-							(property, i) => desiredOrder[i] !== property,
-						)
+						currNode.type === "JSONProperty" &&
+						currNode.key.type === "JSONLiteral"
 					) {
```
JoshuaKGoldberg, line 56: [Bug] What should happen if parent isn't a JSONProperty with a JSONLiteral key? Example:

```js
// eslint.config.js
"sort-collections": ["error", ["gotcha.value"]],
```

```jsonc
// package.json
{
  "gotcha": [
    {
      "value": 123
    }
  ]
}
```

This might actually need a redesign of the options. How do we handle this? Should it be supported at all?

My initial tentative vote: don't support it at first, and consider it an odd edge case? Then if someone files an issue we can always expand the format / add a new allowed one.

JoshuaKGoldberg (changes requested):
🙌 Thanks for sending this in! We normally ask that PRs have associated issues first so we can discuss things there. It's a bit more discoverable and lets folks iron out what scopes of work would make sense to implement. Since this is all working with just a few changes it's fine now, so just a heads up for later.

JoshuaKGoldberg:
> workflow permissions

Yeah, in theory the compliance bot workflow could be fixed to work. But also I've switched a bunch of repos over to https://github.com/JoshuaKGoldberg/OctoGuide that has this issue about forks fixed (and redundant reporting just in case). Filed: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1050

docs/rules/sort-collections.md
andreww2012: Makes sense - should I not add `resolutions` and `dependenciesMeta` too?

src/rules/sort-collections.ts
andreww2012: Currently, the "full" key of the `value` property is going to be `gotcha.value`, as if there are no extra array(s) around `{"value": 123}`.

Should I keep this behavior (which seems fine to me) and if yes, then should I add a corresponding test? This indeed looks like a rare case, and I think the majority of the rule's users won't be affected by that.

JoshuaKGoldberg: Coming back to this, I still am leaning somewhat strongly towards not keeping the behavior. Agreed it's kind of an unusual/weird case that folks aren't likely to hit. But I've seen some _funky_ package.jsons. I think it'd be better to play it safe.

cc @michaelfaith - I could be persuaded if multiple people think it's fine as-is 🙂

michaelfaith: I agree. It feels odd to apply this to array items.

andreww2012: I've removed this behavior, now objects within arrays are unreachable with dot notation, and added corresponding tests.

docs/rules/sort-collections.md
andreww2012: Perhaps you've missed my question, I decided to exclude `resolutions` and `dependenciesMeta` from the defaults 🙂

JoshuaKGoldberg:
😩 what an annoying issue. I'll go ahead and merge #1051 now so we can at least get better reporting of the complaints.

docs/rules/sort-collections.md
JoshuaKGoldberg: Oop yes that, thanks!

PR #1077, opened 2025-06-01
feat(valid-package-definition): add ignoreProperties option

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1076 
- [x] That issue was marked as [`status: accepting prs`](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

This change adds a new `ignoreProperties` option to the `valid-package-definition` rule, allowing you to provide a list of top-level properties to suppress violation reports for.  This can be useful when using this rule alongside any of the more granular `valid-*` rules, so that it's not double reporting violations.

PR #1082, opened 2025-06-01
feat(valid-bin): add new rule for validating bin

### LINKED ISSUE #818: 🚀 Feature: Bring in a valid-bin rule
### Bug Report Checklist

### Overview

Following #42 -> #51: we've resolved to create a standalone `valid-bin` rule. It should enforce that, _if_ the field exists in the `package.json` file, it's set to a valid value. It should do nothing if the field doesn't exist.

### Additional Info

Per discussion in #51, this won't be enabled by default at first. For this issue, we should:

💖

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #818 

## Overview

This change adds a new `valid-bin` rule that uses `validateBin` from `package-json-validator` to identify errors.

This will need to wait to merge after #1077, so that we're not double-reporting validation errors on `bin` with `valid-package-definition`.

Lastly, this is the first new rule that employs our new approach of using the more granular validation functions from `package-json-validator` and can serve as a model for the rest of the `valid-*` rules.

### DIFF
--- a/docs/rules/valid-name.md
+++ b/docs/rules/valid-name.md
@@ -4,7 +4,7 @@
 
 <!-- end auto-generated rule header -->
 
-This rule applies two validations to the `"name"` property:
+This rule applies two validations to the `name` property:
 
 - It must be a string rather than any other data type
 - It should pass [`validate-npm-package-name`](https://www.npmjs.com/package/validate-npm-package-name) validation
--- a/src/rules/valid-bin.ts
+++ b/src/rules/valid-bin.ts
@@ -1,0 +1,46 @@
+import * as ESTree from "estree";
+import { AST as JsonAST } from "jsonc-eslint-parser";
+import { validateBin } from "package-json-validator";
+
+import { createRule } from "../createRule.js";
+
+export const rule = createRule({
+	create(context) {
+		return {
+			"Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.value=bin]"(
+				node: JsonAST.JSONProperty,
+			) {
+				const binValueNode = node.value as unknown as ESTree.Node;
+				const binValue: unknown = JSON.parse(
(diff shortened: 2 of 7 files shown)


docs/rules/valid-name.md
```diff
@@ -4,7 +4,7 @@
 
 <!-- end auto-generated rule header -->
 
-This rule applies two validations to the `"name"` property:
+This rule applies two validations to the `name` property:
```
JoshuaKGoldberg, line 7: [Praise] Oh, nice catch...

src/rules/valid-bin.ts
```diff
@@ -0,0 +1,46 @@
...
+				);
+
+				const errors = validateBin(binValue);
+				if (!errors.length) {
+					return;
+				}
+
+				context.report({
+					data: {
+						errors: errors.join("\n             "),
```
JoshuaKGoldberg: [Feature] This is more spacing in a `.join("\n ... ")` than I've ever seen in a lint rule report before 😄. It's to match up with `"Invalid bin: "` right? If so: I think it's going to be confusing and/or hard to read for users if rules indent their multi-line messages based on the property name.

Proposal: how about switching to a list in the case of multiple errors?

```suggestion
						errors:
							errors.length > 1
								? ["", ...errors].join("\n - ")
								: errors[0],
```

src/tests/rules/valid-bin.test.ts
JoshuaKGoldberg, line 1: [Testing] Missing test case: >1 elements in `errors`. E.g. `"bin": { "": "" }`.

michaelfaith: Updated

src/rules/valid-bin.ts
michaelfaith: Updated

PR #1097, opened 2025-06-09
feat: remove support for node 18

### LINKED ISSUE #1074: Drop support for node 18?
I noticed some renovate updates have started to fail due to those packages dropping support for node 18 (e.g. https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1043).  Thoughts on us dropping support too?  

Alternatively, we could just bump our ci to v20, and retain support for v18.  Though, that becomes risky, because we're not validating that we're not breaking anything for v18, if we do that.

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1074

## Overview

This change removes official support for node v18.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
 	},
 	"packageManager": "pnpm@10.11.0",
 	"engines": {
-		"node": ">=20"
+		"node": "^=20.19.0 || >=22.12.0"
 	},
 	"publishConfig": {
 		"provenance": true
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
 	},
 	"packageManager": "pnpm@10.11.0",
 	"engines": {
-		"node": ">=18"
+		"node": ">=20"
 	},
 	"publishConfig": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -107,7 +107,7 @@
 	},
 	"packageManager": "pnpm@10.11.0",
 	"engines": {
-		"node": ">=18"
+		"node": ">=20"
```
JoshuaKGoldberg: Pinning to versions with just require(ESM) would mean we can assume end-users have that. I think that's part of the intent for #1074 right?

```suggestion
		"node": "^=20.19.0 || >=22.12.0"
```

JoshuaKGoldberg (approved):
Super! Just one suggestion on the minimum engine range for require(ESM). 🔥 

Also ACK that `.nvmrc` is already on 22.

package.json
michaelfaith: Updated

PR #1100, opened 2025-06-10
feat: migrate to ESM-only

### LINKED ISSUE #1080: 🛠 Tooling: switch to bundling our builds
### Tooling Report Checklist

### Overview

Coming from this bug fix: https://github.com/JoshuaKGoldberg/package-json-validator/pull/258  I would propose we shift to bundling our builds like we did for the eslint fix utils package.  There's a known issue with how we're currently doing esm / cjs dual bundling, with our current set up.  Switching to bundling it will resolve that issue, and prevent future regressions like the one fixed in #258.  

### Additional Info

If we don't switch to bundling, then we should at least add an eslint rule to enforce explicit extensions on import paths.

Similar issue opened in `package-json-validator`: https://github.com/JoshuaKGoldberg/package-json-validator/issues/259

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1080

## Overview

This change migrates the package to esm only, and drops dual bundling.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -10,30 +10,11 @@
 		"pnpm-lock.yaml"
 	],
 	"words": [
-		"Codecov",
 		"altano",
-		"autofixable",
-		"codespace",
-		"commitlint",
-		"contributorsrc",
-		"conventionalcommits",
-		"eslint-doc-generatorrc",
-		"esnext",
-		"estree",
-		"knip",
-		"lcov",
-		"markdownlintignore",
+		"generatorrc",
 		"nodenext",
-		"npmpackagejsonlintrc",
-		"outro",
-		"packagejson",
 		"postbuild",
 		"postwatch",
-		"quickstart",
-		"ruleset",
-		"tsup",
-		"wontfix",
-		"zetlan",
 		"zetlen"
 	]
 }
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
 	},
 	"license": "MIT",
 	"author": {
-		"name": "James Zetlan",
-		"email": "zetlan@gmail.com"
+		"name": "James Zetlen",
+		"email": "zetlen@gmail.com"
 	},
 	"type": "module",
 	"exports": {


cspell.json
```diff
@@ -18,10 +18,12 @@
 		"contributorsrc",
 		"conventionalcommits",
 		"eslint-doc-generatorrc",
+		"esnext",
 		"estree",
 		"knip",
 		"lcov",
 		"markdownlintignore",
+		"nodenext",
```
michaelfaith, line 15: FWIW, I feel like the only time I actually interact with cspell, it's always excepting false positives.  I don't think I've actually had it provide any real value. 😓

src/rules/no-empty-fields.ts
```diff
@@ -6,7 +6,7 @@ import {
 } from "eslint-fix-utils";
 import * as ESTree from "estree";
 
-import { createRule, PackageJsonRuleContext } from "../createRule";
+import { createRule, PackageJsonRuleContext } from "../createRule.js";
```
michaelfaith, line 9: This was an actual bug 😊

tsconfig.json
```diff
@@ -3,8 +3,8 @@
 		"declaration": true,
 		"declarationMap": true,
 		"esModuleInterop": true,
-		"module": "ESNext",
-		"moduleResolution": "Bundler",
+		"module": "nodenext",
+		"moduleResolution": "nodenext",
```
michaelfaith, line 7: Using this for `moduleResolution` is sufficient to enforce explicit extensions, without an additional eslint rule.  `Bundler` is intended for packages that actually bundle their code, which is why we kept bumping into this.

JoshuaKGoldberg: Oh nice. Do you think this should be upstreamed to https://github.com/JoshuaKGoldberg/create-typescript-app then?

cspell.json
JoshuaKGoldberg: Interesting, I've felt that it's gotten quite a few of my mistakes anecdotally. I'll keep an eye out for a bit. 

Tangent: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1104 -> https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1105 is an example of a place that it _should_ have caught me, but I didn't listen to it. 😅 

As for these two words, both should already be in the cspell dictionaries:

* `esnext`: https://github.com/streetsidesoftware/cspell-dicts/blob/d803864add6015fed9c5a4b0e0e675eb3a1f88a7/dictionaries/fullstack/src/fullstack.txt#L121 -> https://github.com/streetsidesoftware/cspell-dicts/blob/be7fdb31f9e65f76b5410d353e24ac775631cf8a/dictionaries/typescript/dict/typescript.txt#L1183. This doesn't get flagged for me as a missing word locally.
* `nodenext`: https://github.com/streetsidesoftware/cspell-dicts/pull/4095 only added it to `fullstack`, not `typescript`. Sent: https://github.com/streetsidesoftware/cspell-dicts/pull/4539

So I think this is a transient bug in cspell's dictionaries.

JoshuaKGoldberg (approved):
Awesome!

tsconfig.json
michaelfaith: Yeah, if CTA is prioritizing not bundling, then nodenext would be the better option.

JoshuaKGoldberg: Oh - I just checked and it's already using nodenext. https://github.com/JoshuaKGoldberg/create-typescript-app/blob/afc398b0f27c6b6e8699e7e0d471c64438bbac04/src/blocks/blockTypeScript.ts#L155-L156. Great.

PR #1121, opened 2025-06-17
refactor: augment estree module

### LINKED ISSUE #125: 🛠 Tooling: Find a way to avoid 'as unknown as ESTree.Node' assertions
### Bug Report Checklist

### Overview

Following #40 & #59 & https://github.com/ota-meshi/jsonc-eslint-parser/issues/184: the current state of types for AST nodes in this plugin is that we've completely swapped out the built-in ESLint ones. Nodes are instead typed using `import type { AST as JsonAST } from "jsonc-eslint-parser";` -> `node: JsonAST.JSONProperty` and the like.

Unfortunately, ESLint's APIs still expect ESTree nodes. So we end up with assertions any time we need to pass the nodes to ESLint's APIs:

https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/84035c3f2bac376567eef767ba21f354a0d93f54/src/rules/valid-repository-directory.ts#L32-L34

This isn't ideal. One workaround done by `eslint-plugin-jsonc` is to [instead pass `loc: node.loc`](https://github.com/ota-meshi/eslint-plugin-jsonc/blob/bb0cc408c928ef12952d140add87f3c2b8a2eb40/lib/rules/no-bigint-literals.ts#L26-L27). But I'd ideally like to be able to just pass the `node` directly.

### Additional Info

I don't know a straightforward way to make this better. `context.report` is typed in `@types/eslint` as:

```ts
report(descriptor: ReportDescriptor): void;
```

...with:

```ts
type ReportDescriptor = ReportDescriptorMessage & ReportDescriptorLocation & ReportDescriptorOptions;
type ReportDescriptorMessage = { message: string } | { messageId: string };
type ReportDescriptorLocation =
    | { node: ESTree.Node }
    | { loc: AST.SourceLocation | { line: number; column: number } };
```

I suppose `@types/eslint` could mark the context with a type parameter like `<Node = ESTree.Node>` 😬... Eww.

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #125

## Overview

This change adds the `JSONNode` to `estree`'s `NodeMap`, so that we don't have to clumsily caste AST types from `jsonc-eslint-parser` to `estree` types.  Really, this is something that `jsonc-eslint-parser` should be doing on their side...

Note: we still have to do some clumsy casting for things we pass into the `eslint-fix-utils` functions.  So there could improvements we might need to make on that side.

### DIFF
--- a/src/types/estree.d.ts
+++ b/src/types/estree.d.ts
@@ -1,0 +1,7 @@
+import { JSONNode } from "jsonc-eslint-parser";
+
+declare module "estree" {
+	interface NodeMap {
+		JSONNode: JSONNode;
+	}
+}
--- a/src/rules/no-empty-fields.ts
+++ b/src/rules/no-empty-fields.ts
@@ -53,7 +53,7 @@ const report = (
 	context.report({
 		data,
 		messageId,
-		node: node as unknown as ESTree.Node,
+		node,
 		suggest: [
 			{
 				fix:
--- a/src/rules/no-redundant-files.ts
+++ b/src/rules/no-redundant-files.ts
@@ -55,7 +55,7 @@ export const rule = createRule({
 				context.report({
 					data: { file: element.value },
 					messageId,
-					node: element as unknown as ESTree.Node,
+					node: element,
 					suggest: [
 						{
 							fix: fixRemoveArrayElement(
--- a/src/rules/repository-shorthand.ts
+++ b/src/rules/repository-shorthand.ts
@@ -1,5 +1,3 @@
-import type ESTree from "estree";
-
 import { JSONProperty } from "jsonc-eslint-parser/lib/parser/ast.js";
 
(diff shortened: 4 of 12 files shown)


src/types/estree.d.ts
```diff
@@ -0,0 +1,7 @@
+import { JSONNode } from "jsonc-eslint-parser";
+
+declare module "estree" {
+	interface NodeMap {
+		JSONNode: JSONNode;
+	}
+}
```
JoshuaKGoldberg, line 7: [Praise] _Oh_ I did not at all think of doing this. Very nice, what a clean solution!

Maybe it makes sense to propose adding this to https://github.com/ota-meshi/jsonc-eslint-parser's docs?

JoshuaKGoldberg (approved):
Love it! 🔥

PR #1126, opened 2025-06-19
ci: run tests on all supported node versions

michaelfaith:
## PR Checklist

## Overview

This change updates our ci workflow to run the test stage in CI against node 20, 22, and 24.  This ensures that all supported versions of node work as expected.

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -2,13 +2,19 @@ description: Prepares the repo for a typical CI job
 
 name: Prepare
 
+inputs:
+  node-version:
+    description: "Node.js version to use"
+    default: "20"
+    required: false
+
 runs:
   steps:
     - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4
     - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
       with:
         cache: pnpm
-        node-version: 20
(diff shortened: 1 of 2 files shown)


eslint.config.ts
```diff
@@ -99,10 +99,6 @@ export default tseslint.config(
 		files: ["**/*.{yml,yaml}"],
 		rules: {
 			"yml/file-extension": ["error", { extension: "yml" }],
-			"yml/sort-keys": [
-				"error",
-				{ order: { type: "asc" }, pathPattern: "^.*$" },
-			],
```
JoshuaKGoldberg, line 105: 😄 I've been thinking of getting rid of this in CTA. I like it but recognize it's pretty low impact & takes up space...

michaelfaith: My main gripe with the rule is that it blindly forces an ordering that can actually hurt developer experience more than it helps. The action / workflow yamls are an example. I always want the name at the top followed by triggers (i.e. _what_ is this and _how_ is it used) before the rest of the actual workflow steps. There's a relationship between structure and meaning when it comes to yml files that a pure lexicographic sort can't preserve.

JoshuaKGoldberg: Heh, I don't like having to use any orderings other than alphabetical/lexicographic. There's an unbounded set of reasonable ways to order things like yaml. I personally can never keep track of them in my head. Putting name, then triggers, then the rest is one semi-arbitrary preference that I used to have myself, then would often slip up on.

PR #1158, opened 2025-07-04
feat: add `ignorePrivate` option to all `require-*` rules

### LINKED ISSUE #1092: 🚀 Feature: add an option for `require-name` and `require-version` to ignore package.json's with `private: true` (or make this default)
### Feature Request Checklist

### Overview

[npm documentation says](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#name:~:text=name-,If%20you%20plan%20to%20publish%20your%20package%2C%20the%20most%20important%20things,to%20publish%20your%20package%2C%20the%20name%20and%20version%20fields%20are%20optional.,-The%20name%20is) that "if you don't plan to publish your package, the name and version fields are optional". Usually an indicator of such intention is specifying `"private": true` in package.json. The user that would like to use these rules in say monorepos should manually create a separate flat config and provide a list of package.json's that are intended to be published. I think this is something that should be automated: I'm proposing a new option, say `ignorePrivate`, that should be set to `true` by default for `require-name` and `require-version` rules.

### Additional Info

_No response_

andreww2012:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1092

## Overview

This PR adds `ignorePrivate` option to all `require-*` rules to skip package.json files with `private: true`. It's set to `true` by default only for `require-{name,version}` rules.

@JoshuaKGoldberg [suggested introducing](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1092#issuecomment-2976567386) a global option to control the behavior of the rules in case of private packages, but I'm not too sure how it would work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,11 +130,10 @@ export default {
 
 #### `enforceForPrivate`
 
-Whether `require-*` rules, if used, should enforce the presence of the corresponding property in package.json files with `"private": true`.
-Possible values are:
+**Type:** `boolean`
 
-- `boolean`
-- Not set (default): all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
+Determines whether `require-*` rules, if used, should enforce the presence of the corresponding property in package.json files with `"private": true`.
+By default, all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
 
 ### Usage Alongside Prettier
 
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ Whether `require-*` rules, if used, should enforce the presence of the correspon
 Possible values are:
 
 - `boolean`
-- `"recommended"` (default): all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
+- Not set (default): all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
 
 ### Usage Alongside Prettier
 
--- a/README.md
+++ b/README.md
@@ -101,6 +101,41 @@ module.exports = {
 };
 ```
 
+### Settings
+
+Some rules can be configured in ESLint shared settings.
+You can set them in `settings.packageJson` in an ESLint flat config.
+
+Example:
+
+```ts
+// eslint.config.ts
+import packageJson from "eslint-plugin-package-json";
+
+export default {
+	plugins: {
+		"package-json": packageJson,
+	},
+	rules: {
+		// `description` won't be required in package.json with `"private": true`
+		"package-json/require-description": "error",
+	},
+	settings: {
+		packageJson: {
+			enforceForPrivate: false,
+		},
+	},
+};
+```
+
+#### `enforceForPrivate`
+
+Whether `require-*` rules, if used, should enforce the presence of the corresponding property in package.json files with `"private": true`.
+Possible values are:
+
+- `boolean`
+- `"recommended"` (default): all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
+
 ### Usage Alongside Prettier
 
 **[`prettier-plugin-packagejson`](https://github.com/matzkoh/prettier-plugin-packagejson)** is a [Prettier plugin](https://prettier.io/docs/en/plugins) that enforces the same `package.json` keys ordering as the [`order-properties`](docs/rules/order-properties.md) and [sort-collections](docs/rules/sort-collections.md) rules with default options.
--- a/src/rules/require-properties.ts
+++ b/src/rules/require-properties.ts
@@ -18,7 +18,7 @@ const properties: {
 	{ name: "keywords" },
 	{
 		name: "name",
-		options: { isOptionalForPrivatePackages: true, isRecommended: true },
+		options: { ignorePrivateDefault: true, isRecommended: true },
 	},
 	{ name: "optionalDependencies" },
 	{ name: "peerDependencies" },
@@ -26,7 +26,7 @@ const properties: {
 	{ name: "types" },
 	{
 		name: "version",
-		options: { isOptionalForPrivatePackages: true, isRecommended: true },
+		options: { ignorePrivateDefault: true, isRecommended: true },
 	},
 ];
 
--- a/src/rules/require-properties.ts
+++ b/src/rules/require-properties.ts
@@ -3,35 +3,27 @@ import {
 	createSimpleRequirePropertyRule,
 } from "../utils/createSimpleRequirePropertyRule.js";
 
-const properties: {
-	name: string;
-	options?: CreateRequirePropertyRuleOptions;
-}[] = [
-	{ name: "author" },
-	{ name: "bugs" },
-	{ name: "bundleDependencies" },
-	{ name: "dependencies" },
-	{ name: "description", options: { isRecommended: true } },
-	{ name: "devDependencies" },
-	{ name: "engines" },
-	{ name: "files" },
-	{ name: "keywords" },
-	{
-		name: "name",
-		options: { ignorePrivateDefault: true, isRecommended: true },
-	},
-	{ name: "optionalDependencies" },
-	{ name: "peerDependencies" },
-	{ name: "type", options: { isRecommended: true } },
-	{ name: "types" },
-	{
-		name: "version",
-		options: { ignorePrivateDefault: true, isRecommended: true },
-	},
-];
+const properties: [name: string, options?: CreateRequirePropertyRuleOptions][] =
+	[
+		["author"],
+		["bugs"],
+		["bundleDependencies"],
+		["dependencies"],
+		["description", { isRecommended: true }],
+		["devDependencies"],
+		["engines"],
+		["files"],
+		["keywords"],
+		["name", { ignorePrivateDefault: true, isRecommended: true }],
+		["optionalDependencies"],
+		["peerDependencies"],
+		["type", { isRecommended: true }],
+		["types"],
+		["version", { ignorePrivateDefault: true, isRecommended: true }],
+	];
 
 export const rules = Object.fromEntries(
-	properties.map(({ name: propertyName, options }) => [
+	properties.map(([propertyName, options]) => [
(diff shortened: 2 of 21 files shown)


src/utils/createRequirePropertyRule.ts
```diff
@@ -3,20 +3,48 @@ import type { AST as JsonAST } from "jsonc-eslint-parser";
 import { createRule } from "../createRule.js";
 import { isJSONStringLiteral } from "./predicates.js";
 
+export interface CreateRequirePropertyRuleOptions {
+	isOptionalForPrivatePackages?: boolean;
+	isRecommended?: boolean;
+}
+
+type Options = [{ ignorePrivate?: boolean }?];
```
JoshuaKGoldberg: > @JoshuaKGoldberg [suggested introducing](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1092#issuecomment-2976567386) a global option to control the behavior of the rules in case of private packages, but I'm not too sure how it would work.

I was thinking [Configuring Shared Settings](https://eslint.org/docs/latest/use/configure/configuration-files#configuring-shared-settings). Example prior art: https://perfectionist.dev/guide/getting-started#settings

Adding granular options to rules is a lot of work if you want to set it en masse! A large `private: true` monorepo right now would have to configure this something like a half-dozen times in their config:

```js
import packageJson from "eslint-plugin-package-json";

export default [
  packageJson.configs.recommended,
  {
    rules: {
      "package-json/require-author": [
        "error",
        {
          ignorePrivate: true,
        },
      ],
      "package-json/require-description": [
        "error",
        {
          ignorePrivate: true,
        },
      ],
      "package-json/require-engines": [
        "error",
        {
          ignorePrivate: true,
        },
      ],
      // ...
    },
  },
];
```

Plus, they'd have to know to adjust their config each time we add any new rules that use this too.

Having a single shared setting would save them from that:

```js
import packageJson from "eslint-plugin-package-json";

export default [
  packageJson.configs.recommended,
  {
    settings: {
      packageJson: {
        private: true, // (or something like this)
      },
    },
  },
];
```

Does this make sense? IMO it's a bit of a must-have for ergonomics.

src/utils/createRequirePropertyRule.ts
```diff
@@ -3,20 +3,48 @@ import type { AST as JsonAST } from "jsonc-eslint-parser";
 import { createRule } from "../createRule.js";
 import { isJSONStringLiteral } from "./predicates.js";
 
+export interface CreateRequirePropertyRuleOptions {
+	isOptionalForPrivatePackages?: boolean;
```
michaelfaith: So, when I first saw the name of this prop, where you're using it in `require-properties` (before getting to this code), by the name of it, I assumed it was going to add the `ignorePrivate` option to _only_ the rules with value `true`.  But now that I see how it's used here, it's really just setting what the default value of `ignorePrivate`.   So, two questions come from that experience:

1. Should `ignorePrivate` be available on _all_ `require-*` rules?  As in, does it even make sense for us to have this bail out for every require rule?
2. If yes, and it does make sense for all rules to have the option available, then perhaps, for clarity, the name of this prop should be something like `ignorePrivateDefaultValue`?

michaelfaith: It would also be nice to have TsDoc on each option defined in the interface.

src/utils/createRequirePropertyRule.ts
michaelfaith: A couple questions about this proposal:

1. With some rules `name` and `version` defaulting to `true` and the rest not,  what would the behavior of the global option be?  Especially in the case of a `false` value?  
2. If we do have a global option, would we still have the option at the rule level?  I think we still need it at the rule level.
3. If we did go with this, I'd recommend a slightly different name for the global option.  Just `private` feels very ambiguous to me, and may not be clear, at a glance what its purpose is.  Maybe something like, `relaxForPrivate` or `dontRequireForPrivate` (ew, but also, very clear haha - I'll think some more on it)

michaelfaith (commented):
Great work on this.  Just one additional bit of feedback from me, and comment on @JoshuaKGoldberg's thread.

src/utils/createRequirePropertyRule.ts
JoshuaKGoldberg: 1. I think the global option's default doesn't have to be the same as the rules'. Maybe the global option could be something like `true | false | 'recommended'`? Or some better-named string literals union?
2. Agreed.
3. 👍

michaelfaith: Sounds good. And I wonder if naming and clarity might come easier if we reversed the boolean at the global level. Like `enforceForPrivate`?

src/utils/createRequirePropertyRule.ts
andreww2012: > Should `ignorePrivate` be available on _all_ `require-*` rules?  As in, does it even make sense for us to have this bail out for every require rule?

1. I think it wouldn't hurt to keep this option for all `require-*` rules for the sake of consistency and flexibility.

2. Renamed `isOptionalForPrivatePackages` to `ignorePrivateDefault` instead of `ignorePrivateDefaultValue` because I think this can be misunderstood as "ignore the default value which is private". Added JSDoc too so even if `ignorePrivateDefault` is misunderstood the meaning could be clarified.

src/utils/createRequirePropertyRule.ts
andreww2012: Thanks with designing the shared setting 😄 In "I'm not too sure how it would work" I was referring to what values should this setting accept and how it should be named.

src/rules/require-properties.ts
```diff
@@ -1,30 +1,32 @@
...
-export const rules = properties.reduce<Record<string, PackageJsonRuleModule>>(
-	(acc, [propertyName, isRecommended]) => {
-		acc[`require-${propertyName}`] = createSimpleRequirePropertyRule(
-			propertyName,
-			isRecommended,
-		);
-		return acc;
+const properties: {
+	name: string;
+	options?: CreateRequirePropertyRuleOptions;
```
michaelfaith: Could we model this after `valid-properties`?  (That didn't exist when you first authored this PR, but has since been merged).  It has a very similar purpose, but managed this properties array in a slightly more ergonomic way, where name is still the first element of each array.

src/utils/createSimpleRequirePropertyRule.ts
```diff
@@ -37,13 +77,25 @@ export const createSimpleRequirePropertyRule = (
 		},
 		meta: {
 			docs: {
-				description: `Requires the \`${propertyName}\` property to be present.`,
+				description:
+					"Requires the `${propertyName}` property to be present.",
```
michaelfaith: [Bug] This lost the interpolation for `propertyName`

src/utils/createSimpleRequirePropertyRule.ts
```diff
@@ -11,14 +25,40 @@ import { isJSONStringLiteral } from "./predicates.js";
...
+		ignorePrivateDefault = false,
+		isRecommended,
+	}: CreateRequirePropertyRuleOptions = {},
 ) => {
-	return createRule({
+	return createRule<Options>({
 		create(context) {
+			const enforceForPrivate =
+				context.settings.packageJson?.enforceForPrivate ??
+				"recommended";
```
michaelfaith: I'm not sure I understand why `enforceForPrivate` needs to have a `"recommended"` state.  If we have `undefined` (i.e. user didn't specify the setting), `true` (they want it enabled on everything), and `false` (they don't want it enabled), why do we need another `string` state?

michaelfaith (changes requested):
Good progress!  A few bits of feedback.  Looks very close.

src/utils/createSimpleRequirePropertyRule.ts
andreww2012: I agree. Initially I followed Josh's suggestion, but I think this is actually superfluous.

src/utils/createSimpleRequirePropertyRule.ts
andreww2012: 🤦‍♂️ thanks!

README.md
```diff
@@ -101,6 +101,41 @@ module.exports = {
...
+};
+```
+
+#### `enforceForPrivate`
+
+Whether `require-*` rules, if used, should enforce the presence of the corresponding property in package.json files with `"private": true`.
+Possible values are:
+
+- `boolean`
+- `"recommended"` (default): all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
```
michaelfaith: This documentation still reflects the old shape of the setting.

michaelfaith (changes requested):
Great progress!  Just one more bit of feedback on the documentation of the new setting.

README.md
```diff
@@ -101,6 +101,41 @@ module.exports = {
...
+};
+```
+
+#### `enforceForPrivate`
+
+Whether `require-*` rules, if used, should enforce the presence of the corresponding property in package.json files with `"private": true`.
+Possible values are:
+
+- `boolean`
+- Not set (default): all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
```
JoshuaKGoldberg: [Docs] I found it a bit confusing to have the default behavior explanation after the `boolean`. Also, "values" would mean `true` or `false` I think. How about something like this?

```suggestion
- `false` (default): all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
- `true`: those rules will no longer report if the corresponding property is missing in package.json with `"private": true`.
```

JoshuaKGoldberg (approved):
Wonderful! ✨ 

I just have the one docs suggestion, but it's definitely non-blocking and not urgent. If you disagree with it I won't be offended at all.

As an aside, all the similar changes here make me wish we had some kind of website / system to automate it...

README.md
andreww2012: I've updated the docs in a slightly different way because not setting the value actually differs from `false` or `true`.

JoshuaKGoldberg (approved):
Looks great to me! @michaelfaith was also touching this area so there might be some merge conflicts, just a heads up.

michaelfaith:
I resolve the conflicts and added the tests back to the new `require-properties` file.   It looks like the new option just needs to be added to the docs for the new `require-` rules that have been added recently.  It would also be really great if you could add `name` properties to all the tests.  It's really hard to identify which tests are failing when they do, by just the text of the code, since the differences between tests are in the value of their options (and the plugin settings).  

image

@JoshuaKGoldberg I wonder if we should enforce that in some way via lint rule? 🤔

michaelfaith:
And I'll refrain from adding more `require-` rules til this merges.  Sorry for the additional headache.

michaelfaith:
I added the missing docs

PR #1185, opened 2025-07-24
refactor: add utility function for generating basic `valid-` rules

michaelfaith:
## PR Checklist

## Overview

This change adds a new utility function in the same vein as the one we use to generate simple `require-` rules, but for simple `valid-` rules.  If the valid- rule doesn't need any options, and is simply using `package-json-validator` to do the validation, then it can be generated using this new function.


michaelfaith:
Code coverage dropped because I deleted a number of files that were 100% coverage.  No new lines of code are uncovered.

README.md
```diff
@@ -137,9 +137,9 @@ The default settings don't conflict, and Prettier plugins can quickly fix up ord
 | [restrict-dependency-ranges](docs/rules/restrict-dependency-ranges.md) | Restricts the range of dependencies to allow or disallow specific types of ranges.                |      |    | 💡 |    |
 | [sort-collections](docs/rules/sort-collections.md)                     | Dependencies, scripts, and configuration values must be declared in alphabetical order.           | ✔️ ✅ | 🔧 |    |    |
 | [unique-dependencies](docs/rules/unique-dependencies.md)               | Checks a dependency isn't specified more than once (i.e. in `dependencies` and `devDependencies`) | ✔️ ✅ |    | 💡 |    |
-| [valid-author](docs/rules/valid-author.md)                             | Enforce that the author field is a valid npm author specification                                 | ✔️ ✅ |    |    |    |
+| [valid-author](docs/rules/valid-author.md)                             | Enforce that the `author` property is valid.                                                      | ✔️ ✅ |    |    |    |
```
JoshuaKGoldberg, line 140: 😕 [Bug] This is a loss of functionality IMO. The current/previous description is a little more informative (though the backticks around the name are definitely an improvement). How about allowing an optional description for the utility function?

src/rules/valid-properties.ts
JoshuaKGoldberg, line 1: [Naming] Nit: maybe, `valid-property-rules.ts`? I was confused at first seeing an import of `rules` (plural) from a file with the same name format as other rule files.

src/rules/valid-properties.ts
```diff
@@ -0,0 +1,59 @@
...
+		}
+		acc[`valid-${propertyName}`] = createSimpleValidPropertyRule(
+			propertyName,
+			validationFunction,
+			aliases,
+		);
+		return acc;
+	},
+	{},
+);
```
JoshuaKGoldberg, line 59: [Refactor] I personally find the combo of `const [ ... ] = ... ? ... : ...` pattern and the `Object.fromEntries` a little easier to read. Absolutely not blocking though, just a mild suggestion.

```suggestion
export const rules = Object.fromEntries(
	properties.map(([propertyName, validationFunctionOrOptions]) => {
		const [aliases, validationFunction] =
			typeof validationFunctionOrOptions === "object"
				? [
						validationFunctionOrOptions.aliases,
						validationFunctionOrOptions.validator,
					]
				: [[], validationFunctionOrOptions];

		return [
			`valid-${propertyName}`,
			createSimpleValidPropertyRule(
				propertyName,
				validationFunction,
				aliases,
			),
		];
	}),
);
```

src/utils/createSimpleValidPropertyRule.ts
```diff
@@ -0,0 +1,61 @@
...
+	propertyName: string,
+	validationFunction: ValidationFunction,
+	aliases: string[] = [],
+) => {
+	const propertyNames = [propertyName, ...aliases];
+	return createRule({
+		create(context) {
+			return propertyNames.reduce<
+				Record<string, (node: JsonAST.JSONProperty) => void>
+			>((acc, name) => {
```
JoshuaKGoldberg, line 24: [Style] Similar very-much-not-a-blocker suggestion here for a less object-mutating-y pattern:

```ts
			return Object.fromEntries(
				propertyNames.map((name) => [
					`Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.value=${name}]`,
```

JoshuaKGoldberg (approved):
Nice! +141 -299 is wonderful. I only have a few stylistic suggestions that are functionally equivalent, no actual blockers. ✨

README.md
michaelfaith: Oh, the original version doesn't seem clear at all to me (the npm author doesn't have a spec.  npm has a spec that includes an author field haha), and across all the valid- rules, we had slightly different versions of that text (I blame me for that 😓!).   I guess, why would we want to have that be different for any of them?  Happy to change the text to something you think is more descriptive, but I think we'd want them to all be the same, right?

src/rules/valid-properties.ts
michaelfaith: I was just matching the `require-properties` naming from the other implementation.  I'm happy to change both to that.

src/rules/valid-properties.ts
michaelfaith: Sure, I can change that, but it's just using the same approach as I did in `require-properties`.  so if we change it here, it should be changed there too for consistency.  I actually don't find the Object.entries / Object.from approach easier to reason about.  🤷

michaelfaith: I think memory and speed characteristics are (minuscule-y) better for the reduce approach.  Unless you feel strongly about it, I'm inclined to leave the reduce versions of these?

JoshuaKGoldberg: I think they'll probably be equivalent at the scale this linter operates at. But yeah, +1 to consistency. This is a very low priority change suggestion for me 😄

src/rules/valid-properties.ts
JoshuaKGoldberg: I don't have a strong preference here, whatever you like more 🙂

README.md
JoshuaKGoldberg: Hmm now that you mention it, the original one is a little unclear... yeah, let's just go with consistency. 👍

PR #1218, opened 2025-08-11
feat: add docs url to all rules

### LINKED ISSUE #1217: 🚀 Feature: add docs url to rules' meta
### Feature Request Checklist

### Overview

It would be really handy to be able to click the rule name to quickly browse its documentation right in the editor:

Missing docs link when hover over a problem in the editor coming from eslint-plugin-package-json

### Additional Info

_No response_

andreww2012:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1217

## Overview

This PR changes the signature of `createRule` which now requires rule name to be passed in order to generate the link to the rule docs.

### DIFF
--- a/src/createRule.ts
+++ b/src/createRule.ts
@@ -50,16 +50,22 @@ export interface PackageJsonSourceCode extends SourceCode {
 }
 
 export function createRule<Options extends unknown[]>(
-	rule: PackageJsonRuleModule<Options>,
-) {
+	rule: PackageJsonRuleModule<Options> & { name: string },
+): PackageJsonRuleModule<Options> {
 	return {
-		...rule,
 		create(context: PackageJsonRuleContext<Options>) {
 			if (!isPackageJson(context.filename)) {
 				return {};
 			}
 
 			return rule.create(context);
+		},
+		meta: {
+			...rule.meta,
+			docs: {
+				...rule.meta.docs,
+				url: `https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/HEAD/docs/rules/${rule.name}.md`,
+			},
 		},
 	};
 }
--- a/src/rules/valid-package-definition.ts
+++ b/src/rules/valid-package-definition.ts
@@ -47,7 +47,6 @@ export const rule = createRule<Options>({
 			},
 		};
 	},
-
 	// eslint-disable-next-line eslint-plugin/prefer-message-ids
 	meta: {
 		defaultOptions: [{ ignoreProperties: [] }],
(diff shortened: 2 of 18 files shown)


src/rules/valid-package-definition.ts
```diff
@@ -38,7 +38,6 @@ export const rule = createRule<Options>({
 				for (const error of usableErrors) {
 					if (error.message) {
 						context.report({
-							// eslint-disable-next-line eslint-plugin/prefer-message-ids
```
michaelfaith: Hmm, why did these disables go away?  They're still outstanding issues?  Did changing the rule to the second parameter in the function throw off `eslint-plugin-eslint-plugin`'s rule detection?

src/createRule.ts
```diff
@@ -61,5 +62,12 @@ export function createRule<Options extends unknown[]>(
 
 			return rule.create(context);
 		},
+		meta: {
+			...rule.meta,
+			docs: {
+				...rule.meta.docs,
+				url: `https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/HEAD/docs/rules/${ruleName}.md`,
```
michaelfaith: I would almost rather the `url` prop be in each rule file, rather than abstracted here.  Only slightly though.  But if setting it up this way, hampers our ability to do proper plugin linting in the rule definitions, then I'll feel a lot more strongly about it.  Does this impact rule linting?

JoshuaKGoldberg (approved):
Great! I tried it out locally and it works very well. 🚀 thanks!

Aside: as a maybe good big next, I filed #1219

michaelfaith (changes requested):
Sure enough. I downloaded this branch and we're not getting any linting from `eslint-plugin-eslint-plugin` anymore.  

I tried removing `meta.schema` from one rule (which is required by the `recommended` config of that plugin), and it didn't report the violation.  So, we're not going to be able to move forward with it in that state.  I suspect it's dude to the rule object being move to the second param, like I mentioned in my earlier comment.

src/rules/valid-package-definition.ts
andreww2012: Totally missed this one. Thanks! (one of the reasons I'm not a fan of auto-fix on save 😅)

src/createRule.ts
andreww2012: I tried to avoid resorting to URL copy-pasting or introducing a helper to generate such a URL, which would need to be imported in every rule's file.

I now changed the way of passing the rule name to `createRule`, moving it to the object with `create` and `meta`. This also brought back `eslint-plugin-eslint-plugin`.

JoshuaKGoldberg:
Oh interesting, I tried a couple rules out locally and did get errors. But clearly I didn't look deep enough. Thanks!

michaelfaith (approved):
Looks good. Thanks!

PR #1250, opened 2025-09-03
fix: improve sort-collections docs and error messages

### LINKED ISSUE #1243: 📝 Documentation: `sort-collections` doesn't fully document the behavior of `scripts` sorting
### Documentation Report Checklist

- [x] I have checked the latest `main` branch of the repository.

### Overview

Coming from https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1203#issuecomment-3246324956, the docs for `sort-collections` doesn't describe the behavior of sorting `scripts` in relation to npm lifecycle hooks (e.g. `postinstall`, `prebuild`, etc.).

### Additional Info

_No response_

KristjanESPERANTO:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #1243
-   [x] Steps in [CONTRIBUTING.md](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/main/.github/CONTRIBUTING.md?rgh-link-date=2025-09-02T22%3A23%3A42Z) were taken

## Overview

Originally, this was only supposed to be a docs PR. However, I thought it would be logical to include the other changes too, since they are somehow related. Let me know if you want PRs for each commit.

### Commits

1. docs(sort-collections): add Scripts Lifecycle Ordering section and update README description
   - Previous docs didn’t reflect lifecycle-aware scripts ordering.
   - Fixes #1243.
2. refactor(sort-collections): generalize error message and unify terminology
  - The previous error message did not accurately describe the problem.
    - Before: `Package scripts are not alphabetized`
    - After: `Entries in 'scripts' are not in the expected order`
  - The terminology in the code was inconsistent and, in some cases, inaccurate.
3. test(valid-local-dependency): rename fixture to `local-dependency-context` for clarity
  - The previous fixture name `unalphabetized-collections` implied ordering relevance and did not contain anything unalphabetized.


src/rules/sort-collections.ts
```diff
@@ -137,12 +137,13 @@ export const rule = createRule<Options>({
...
 			description:
-				"Dependencies, scripts, and configuration values must be declared in alphabetical order.",
+				"Selected package.json collections must be in a consistent order (lexical for most; lifecycle-aware for scripts).",
 			recommended: true,
 		},
 		fixable: "code",
 		messages: {
-			notAlphabetized: "Package {{ key }} are not alphabetized",
+			incorrectOrder:
+				"Entries in '{{ key }}' are not in the expected order",
```
JoshuaKGoldberg: 💡 Idea: maybe there should be a separate message for `scripts` vs. the rest of the fields? Since `scripts` is its own special thing, I think it'd be helpful for disambiguation.

```suggestion
			incorrectOrderAlphabetical:
				"Package '{{ key }}' is expected to be alphabetized",
			incorrectOrderScripts:
				"Package 'scripts' are expected to be alphabetized and grouped by lifecycles",
```

...or maybe even only including the _"`grouped by lifecycles`"_ point for lifecycle scripts? Some better phrasing around that?

I don't think this needs to block this PR. It might be better as a followup issue to discuss more. But just ideating. 🙂

docs/rules/sort-collections.md
```diff
@@ -10,10 +10,12 @@ Whenever npm changes package dependencies through `npm install`, it lexically so
 However, developers will manually update `package.json` and accidentally leave dependencies out of order.
 Doing so leads to "noise" in commits when a later change re-sorts the packages.
 
-Furthermore, the `"scripts"` collections in `package.json` files are generally easiest to work with when in alphabetical order.
-To keep related scripts close together in the file, consider using the common npm script convention of colon-separated namespaces, such as `watch:all` or `watch:dist`.
+For most collections (`dependencies`, `peerDependencies`, `exports` entries, etc.) this rule enforces a simple ascending lexical ordering (based on raw string keys).
 
-This rule aims to keep the dependency collections sorted in every commit.
+`scripts` are an exception: they use a lifecycle‑aware ordering, which groups `pre<name>` / `<name>` / `post<name>` together (even if the middle one is missing).
+Details & examples: see [Scripts Lifecycle Ordering](#scripts-lifecycle-ordering).
```
JoshuaKGoldberg: [Docs] Just nitpicking, this is the only sentence not written like typical conversational flow.

```suggestion
See [Scripts Lifecycle Ordering](#scripts-lifecycle-ordering) for more details and examples.
```

docs/rules/sort-collections.md
```diff
@@ -77,6 +79,55 @@ The following patterns are **not** considered errors:
...
+### Scripts Lifecycle Ordering
+
+Summary:
+
+- Lifecycle scripts form a group: `pre<name>` → `<name>` → `post<name>`.
+- The group is positioned where `<name>` would appear lexically, even if `<name>` itself is missing.
+- Groups with different base names (`build`, `install`, `test`, ...) are ordered relative to each other by their base name.
+- All other script names (including namespaced ones like `lint:fix`, `watch:dist`) then follow the `sort-package-json` ordering after lifecycle grouping.
+
+This matches the behavior of `prettier-plugin-packagejson`.
```
JoshuaKGoldberg:
```suggestion
This matches the behavior of [`prettier-plugin-packagejson`](https://github.com/matzkoh/prettier-plugin-packagejson).
```

JoshuaKGoldberg (approved):
So thorough! Especially for a PR sent half a day after the issue was filed 😄 nicely done.

No blockers from me, just some docs touchups and an idea that might just be better as a followup.

src/rules/sort-collections.ts
KristjanESPERANTO: Good point. I'll take a look at it right away :)

KristjanESPERANTO: I implemented this with the last commit:

- scripts: `Entries in 'scripts' are not lexically sorted and grouped by lifecycles`
- other collections: `Entries in '{{ key }}' are not lexically sorted`

JoshuaKGoldberg (approved):
💯 thanks!

src/rules/sort-collections.ts
```diff
@@ -60,7 +60,7 @@ export const rule = createRule<Options>({
 				const currentOrder = collection.properties;
 				let desiredOrder: JsonAST.JSONProperty[];
 
-				// If it's any property other than `scripts`, simply sort lexicographically
+				// If it's any property other than `scripts`, simply sort lexically
```
michaelfaith: "Lexicographically" was probably the better word to use (here and in the violation messages).  Curious what prompted the change to "lexically"?  https://en.wikipedia.org/wiki/Lexicographic_order

src/rules/sort-collections.ts
```diff
@@ -137,12 +140,15 @@ export const rule = createRule<Options>({
...
 			description:
-				"Dependencies, scripts, and configuration values must be declared in alphabetical order.",
+				"Selected package.json collections must be in a consistent order (lexical for most; lifecycle-aware for scripts).",
 			recommended: true,
 		},
 		fixable: "code",
 		messages: {
-			notAlphabetized: "Package {{ key }} are not alphabetized",
+			incorrectOrderLexical:
+				"Entries in '{{ key }}' are not lexically sorted",
```
michaelfaith: Similar to the above, curious about the use of "lexically" instead of "lexicographically"?

michaelfaith (commented):
Agreed on the thoroughness!  Nice work.  I do have one question about the choice of word change from "lexicographically" which seems more correct to "lexically".

src/rules/sort-collections.ts
KristjanESPERANTO: I'm not a native speaker, and at first I thought "lexically" and "lexicographically" meant the same in this context. I preferred the shorter form for consistency. After looking it up, I understand now that "lexicographically" is the more precise and correct term here. So, I think we should revert that.

KristjanESPERANTO: But today I can not do it.

michaelfaith: Sounds good!  I'll mark as Change Requested til you're able to get back to it.

michaelfaith (changes requested):
Marking as Request Changes per https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1250#discussion_r2320324228

src/rules/sort-collections.ts
KristjanESPERANTO: OK, I've made some changes to avoid using the term 'lexical' :slightly_smiling_face: 

To restate: In practice, the terms 'lexicographical' and 'lexical' appear to be used interchangeably when it comes to sorting. However, this is not strictly correct, so we should avoid using the term 'lexical'.

src/rules/sort-collections.ts
KristjanESPERANTO: Changed. See comment above. :slightly_smiling_face:

src/rules/sort-collections.ts
KristjanESPERANTO: After other change requests it's now:

- scripts: `Entries in 'scripts' are not in lexicographical order and grouped by lifecycles`
- other collections: `Entries in '{{ key }}' are not in lexicographical order`

michaelfaith (approved):
Looks good!  Thanks again

PR #1258, opened 2025-09-11
refactor: infer rule options type from json schema

### LINKED ISSUE #1259: 🔧 Tooling: infer rule options type from JSON schema
### Feature Request Checklist

### Overview

I noticed that rule options currently need to be specified manually:

https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/71f819237dfa9c9eb2ccaa9a5b06e72e16ce5ab9/src/rules/no-empty-fields.ts#L11-L14

However, they must stay in sync with the options JSON schema. Additionally, `meta.defaultOptions` is not constrained to the options type on type level.

There exists a package [json-schema-to-ts](https://www.npmjs.com/package/json-schema-to-ts) which we could leverage to infer the options type automatically and also use the inferred type on `meta.defaultOptions`.

### Additional Info

_No response_

andreww2012:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1259

## Overview

I'm not sure if this PR requires an issue. I'm happy to open one if it is 🙂

I've noticed that rule options types are written manually and thus must be in sync with JSON schema. This is a QoL PR that makes type inference automatic using [`json-schema-to-ts`](https://www.npmjs.com/package/json-schema-to-ts). This change also allowed `meta.defaultOptions` to be typed.

If for some reason `json-schema-to-ts` is not able to correctly infer the type, there's an escape hatch that allows manually passing the options type to the first type parameter of `createRule`.

### DIFF
--- a/src/createRule.ts
+++ b/src/createRule.ts
@@ -63,9 +63,14 @@ type InferJsonSchemasTupleType<T extends JSONSchema[]> = {
 	[K in keyof T]?: InferJsonSchemaType<T[K]>;
 };
 
+/**
+ * Rule options type is inferred from the JSON schema by [json-schema-to-ts](https://www.npmjs.com/package/json-schema-to-ts).
+ * If you're not satisfied with the inferred type, you may specify it manually in the first type parameter.
+ */
 export function createRule<
 	OptionsOverride extends unknown[] = never,
 	const Schema extends JSONSchema[] = JSONSchema[],
+	// `[T] extends [never]` correctly checks if `T` is `never`, see https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types
 	_OptionsResolved extends unknown[] = [OptionsOverride] extends [never]
 		? InferJsonSchemasTupleType<Schema>
 		: OptionsOverride,


pnpm-workspace.yaml
michaelfaith, line 1: Since we're not using workspaces, can we move this to the `package.json`?  We can remove the eslint config change with that shift.

src/createRule.ts
```diff
@@ -40,20 +44,36 @@ export interface PackageJsonRuleContext<Options extends unknown[] = unknown[]>
...
 }
 
 export interface PackageJsonSourceCode extends SourceCode {
 	ast: PackageJsonAst;
 }
 
-export function createRule<Options extends unknown[]>(
-	rule: PackageJsonRuleModule<Options> & { name: string },
-): PackageJsonRuleModule<Options> {
+type InferJsonSchemasTupleType<T extends JSONSchema[]> = {
```
michaelfaith, line 62: Nice work on these types.

michaelfaith (changes requested):
Awesome. Thank you!  Just a couple of minor things.

pnpm-workspace.yaml
andreww2012: Since pnpm@10.15.0 is declared as a `packageManager`, it seems like, [according to the docs](https://pnpm.io/settings), putting all pnpm settings in `pnpm-workspace.yaml` file is a recommended way of configuring it in this version, and creating such a file doesn't necessarily state that the workspaces feature is used. It even was created automatically after first running [`pnpm approve-builds`](https://pnpm.io/cli/approve-builds).

Should I still move everything from this file to `pnpm` key in `package.json` or can I stick with the current solution? It also seems like having `pnpm-workspace.yaml` isn't affecting anything.

JoshuaKGoldberg: Yeah, this and the other config changes feel like something to be split out into a separate PR. They don't need to land with (or be blocked by) the main changes in this PR.

JoshuaKGoldberg (approved):
+1 to what @michaelfaith said, these types are very nifty. Nicely done. 🙂 

My only concern is that these types of types shenanigans can be unapproachable or complex to work with. Lines like `_OptionsResolved extends ... extends [never]` are not something 99% of TypeScript developers would understand. I'm up for trying this out if @michaelfaith is up for it - the inference of options types is quite useful. But something to keep in mind.

pnpm-workspace.yaml
michaelfaith: > Since pnpm@10.15.0 is declared as a `packageManager`, it seems like, [according to the docs](https://pnpm.io/settings), putting all pnpm settings in `pnpm-workspace.yaml` file is a recommended way of configuring it in this version

You're right. I was looking at an older version of the docs.  I'm ok with this.

michaelfaith (approved):
Looks like there's a conflict now with the `createRule` file (after merging an unrelated bug fix).  Besides that, I'm comfortable with moving forward with this change.

andreww2012:
Added some comments.

PR #1266, opened 2025-09-15
docs: update readme

Zamiell:
## PR Checklist

## Overview

When documenting options, the default values should be at the top so that it is easy to see at a glance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.56.3](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/compare/v0.56.2...v0.56.3) (2025-09-15)
+
+### Bug Fixes
+
+- **plugin:** update types to be compatible with `defineConfig` ([#1245](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1245)) ([861e5e1](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/commit/861e5e129b3a5e3455b6b2a2a26d453086f20c14)), closes [#1242](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1242)
+
 ## [0.56.2](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/compare/v0.56.1...v0.56.2) (2025-09-04)
 
 ### Bug Fixes
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "eslint-plugin-package-json",
-	"version": "0.56.2",
+	"version": "0.56.3",
 	"description": "Rules for consistent, readable, and valid package.json files. 🗂️",
 	"homepage": "https://github.com/JoshuaKGoldberg/eslint-plugin-package-json#readme",
 	"bugs": {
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -130,7 +130,8 @@ export default {
 
 #### `enforceForPrivate`
 
-**Type:** `boolean`
+- **Type:** `boolean`
+- **Default:** `true`
```
michaelfaith: The default for this setting is not actually `true`, though.   Hint: it's not `false` either 😆.  Two lines down says what the default behavior is.  Each rule determines what its default behavior is (true or false).  This global setting, overrides those defaults to be either `true` for all, or `false` for all.

JoshuaKGoldberg: Yeah, since the default is not a trivial explanation, I think I prefer it as-is.

Zamiell: Oh, that's very confusing. Let me try to rewrite the section then to make this behavior more clear.

Zamiell: I've rewritten the section now. Even though there is no clear default value, I still prefer clearly calling out what the default value is at the top. It provides a nice summary of what is to come, and it creates a precedent of how the formatting should be if more settings are added in the future.

README.md
```diff
@@ -130,10 +130,19 @@ export default {
 
 #### `enforceForPrivate`
 
-**Type:** `boolean`
+- **Type:** `boolean`
+- **Default:** [dynamic]
```
michaelfaith: Maybe `See below` here instead of `[dynamic]`?

README.md
```diff
@@ -130,10 +130,19 @@ export default {
...
 
-Determines whether `require-*` rules, if used, should enforce the presence of the corresponding property in package.json files with `"private": true`.
-By default, all `require-*` rules except for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md) will report if the corresponding property is missing in package.json with `"private": true`.
+When a package.json file has a `"private": true` field, it indicates that the package will not be published to npm (or another online registry).
+Some fields that are nice to have in public packages become pointless when a package is private.
```
michaelfaith: "pointless" might be a bit strong.  Recommend

```suggestion
Some fields that are nice to have in public packages become less relevant when a package is private.
```

README.md
```diff
@@ -130,10 +130,19 @@ export default {
...
+This option determines whether `require-*` rules, if used, should enforce the presence of the corresponding property in package.json files that have `"private": true`.
+
+By default, this is:
+
+- `false` for [`require-name`](docs/rules/require-name.md) and [`require-version`](docs/rules/require-version.md).
+- `true` for every other `require-*` rule.
+
+By specifying this setting as `true` or `false`, it will override this logic and change the setting for ALL rules. (Meaning that either all `require-*` rules rules will be applied to private packages or no `require-*` rules rules will be applied to private packages, respectively.
```
michaelfaith: This has a missing closing paren and an extra "rules"

```suggestion
By specifying this setting as `true` or `false`, it will override the defaults and apply the setting for ALL rules.  
In that case, either all `require-*` rules will be applied to private packages or no `require-*` rules will be applied to private packages.  
Even then, you can override the setting again at the rule level, by using the rule's `ignorePrivate` option, which will take precedence over this global setting.
```

michaelfaith (changes requested):
I like it.  This is looking much better.  Just a few feedback items from me.

README.md
Zamiell: I will resolve this thread in favor of the other one.

michaelfaith (approved):
Looks good. Thanks again!

PR #1270, opened 2025-09-16
build: replace usage of tseslint.config

### LINKED ISSUE #1244: 🛠 Tooling: replace usages of `tseslint.config`
### Tooling Report Checklist

### Overview

In the latest release of `typescript-eslint` their `config` function was deprecated in favor of ESLint's built-in `defineConfig` function.  We're using the `tseslint` function in our repo's config, and should shift to using `defineConfig` instead.

### Additional Info

_No response_

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1244

## Overview

This change moves our eslint config to use ESLint's `defineConfig` function.  I also removed usage of `eslint-plugin-markdown`, since it's been deprecated (and replaced by `@eslint/markdown` since last year).

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -1,3 +1,4 @@
+// @ts-expect-error - no types
 import comments from "@eslint-community/eslint-plugin-eslint-comments/configs";
 import eslint from "@eslint/js";
 import vitest from "@vitest/eslint-plugin";


docs/rules/order-properties.md
```diff
@@ -17,17 +17,13 @@ This rule detects when properties in `package.json` are out of order.
 
 Examples of **incorrect** code for this rule:
 
-<!-- eslint-disable jsonc/sort-keys -->
```
JoshuaKGoldberg, line 20: Err, isn't removing this an indication that `eslint-plugin-markdown` was doing something no other plugin was? This makes me think `@eslint/markdown` will need to be configured more?

michaelfaith: Just that it provided the markdown processor, and other rules were being applied to markdown (like this json rule).  I would argue that it's doing something we didn't want it to do, since we actively disabled it 😆.

Also, I didn't add `@eslint/markdown` here.  We're using `markdownlint` still, and in my opinion we shouldn't have two competing tools for linting markdown.  The move from `markdownlint` to `@eslint/markdown` should ideally stand on its own as an intentional move, which I believe you've already captured in https://github.com/JoshuaKGoldberg/create-typescript-app/issues/1926

JoshuaKGoldberg: Haha I suppose it is partially what we don't want. But it does also lint code snippets in the README.md, many of the valid cases, and really all the rules except the disabled ones, no?

Anyway this isn't a blocker IMO, just a curiosity. 👍 from me.

michaelfaith:
Rebased to resolve conflict

PR #1328, opened 2025-10-17
feat(exports-subpaths-style): add new rule

### LINKED ISSUE #1322: 🚀 Feature: Add `exports-subpaths-style` rule to control exports field format
## Overview

Add a new rule that enforces consistent formatting for the `exports` field (similar to how `repository-shorthand` works).

### Proposed formatting options
`implicit` (default) or `explicit`

1. **Single root export with string value:**
   ```json5
   // implicit (default)
   "exports": "./a.js"
   
   // explicit
   "exports": { ".": "./a.js" }
   ```

2. **Single root export with conditional exports:**
   ```json5
   // implicit
   "exports": { "import": "./a.js", "require": "./b.js" }
   
   // explicit
   "exports": { ".": { "import": "./a.js", "require": "./b.js" } }
   ```

### When NOT to apply
- Multiple subpath exports: `{ ".": "./index.js", "./util": "./util.js" }` - must use explicit subpaths form

### Configuration

```typescript
{
  "prefer": "implicit" | "explicit"
}
```

- `"implicit"` (default) - omit `"."` key when only exporting root
- `"explicit"` - always include `"."` key to make subpaths structure explicit

### Benefits:
- Enforces team preference for exports formatting
- Reduces nesting when only exporting root (implicit mode)
- Makes subpath structure explicit when desired (explicit mode)
- Consistent with `repository-shorthand` pattern

**Auto-fix:** Should be auto-fixable in both directions

Happy to implement this if approved.

privatenumber:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1322

## Overview

- Added `exports-subpaths-style` rule that enforces consistent format for the `exports` field
  - Supports two modes: `{ "prefer": "explicit" | "implicit" }` with `"explicit"` as default
  - Auto-fixable in both directions
  - Handles string exports: `"./index.js"` <-> `{ ".": "./index.js" }`
  - Handles conditional exports: `{ "import": "...", "require": "..." }` <-> `{ ".": { "import": "...", "require": "..." } }`
  - Skips transformation when multiple subpaths exist (always explicit form)
  - Ignores `null` values and nested exports fields

- [ ] TODO: Currently not included in `recommended` config. Should we add it?

### DIFF
--- a/docs/rules/exports-subpaths-style.md
+++ b/docs/rules/exports-subpaths-style.md
@@ -4,6 +4,134 @@
 
 <!-- end auto-generated rule header -->
 
+This rule enforces consistent formatting for the `exports` field in `package.json`, ensuring either explicit subpath notation with `"."` or implicit root-level exports.
+
+## Rule Details
+
+The Node.js `exports` field supports two equivalent formats for defining a single root export:
+
+**Explicit format** (with `"."` subpath):
+
+```json
+{
+	"exports": {
+		".": "./index.js"
+	}
+}
+```
+
+**Implicit format** (root-level value):
+
+```json
+{
+	"exports": "./index.js"
+}
+```
+
+Both formats are functionally equivalent for a single root export, but using one consistently improves readability and maintainability across packages.
+
+> 💡 This rule is especially useful in monorepos where consistency across multiple `package.json` files is important.
+
+### Examples with Default Options (prefer: "explicit")
+
+Examples of **incorrect** code with default options:
+
+```json
+{
+	"exports": "./index.js"
+}
+```
+
+```json
+{
+	"exports": {
+		"import": "./index.mjs",
+		"require": "./index.cjs"
+	}
+}
+```
+
+Examples of **correct** code with default options:
+
+```json
+{
+	"exports": {
+		".": "./index.js"
+	}
+}
+```
+
+```json
+{
+	"exports": {
+		".": {
+			"import": "./index.mjs",
+			"require": "./index.cjs"
+		}
+	}
+}
+```
+
+```json
+{
+	"exports": {
+		".": "./index.js",
+		"./utils": "./utils.js"
+	}
+}
+```
+
+### Examples with prefer: "implicit"
+
+Examples of **incorrect** code with `{ "prefer": "implicit" }`:
+
+```json
+{
+	"exports": {
+		".": "./index.js"
+	}
+}
+```
+
+```json
+{
+	"exports": {
+		".": {
+			"import": "./index.mjs",
+			"require": "./index.cjs"
+		}
+	}
+}
+```
+
+Examples of **correct** code with `{ "prefer": "implicit" }`:
+
+```json
+{
+	"exports": "./index.js"
+}
+```
+
+```json
+{
+	"exports": {
+		"import": "./index.mjs",
+		"require": "./index.cjs"
+	}
+}
+```
+
+```json
+{
+	"exports": {
+		".": "./index.js",
+		"./utils": "./utils.js"
+	}
+}
+```
+
+Note: Multiple subpaths always require the explicit object format and are not transformed by this rule.
+
 ## Options
 
 <!-- begin auto-generated rule options list -->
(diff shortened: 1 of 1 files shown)


src/rules/exports-subpaths-style.ts
```diff
@@ -0,0 +1,149 @@
...
+		},
+		schema: [
+			{
+				additionalProperties: false,
+				properties: {
+					prefer: {
+						description:
+							"Specifies which exports format to enforce.",
+						enum: ["implicit", "explicit"],
+						type: ["string"],
```
michaelfaith: Nit: this can be a regular string (rather than array). 
```suggestion
						type: "string",
```

src/rules/exports-subpaths-style.ts
```diff
@@ -0,0 +1,149 @@
...
+				return;
+			}
+
+			// Skip null or non-string literals
+			if (node.value.type === "JSONLiteral" &&
+			    (node.value.value === null || typeof node.value.value !== "string")) {
+				return;
+			}
+
+			const exportsValue = node.value as ExportsValue;
```
michaelfaith: Nit: if this were set at the top of the function (without the `as` casting), and then `exportsValue` used in if conditions, would TypeScript properly narrow the type for you, without having to cast?

src/rules/exports-subpaths-style.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+function isImplicitFormat(node: ExportsValue): boolean {
+	if (node.type === "JSONLiteral") {
+		return true;
+	}
+
+	const keys = node.properties
+		.filter((property) => property.key.type === "JSONLiteral" && typeof property.key.value === "string")
+		.map((property) => (property.key as JsonAST.JSONStringLiteral).value);
```
michaelfaith: There's a type predicate function for `JSONStringLiteral` [here](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/main/src/utils/predicates.ts#L3-L7) that should help narrow the type and avoid the casting.

src/rules/exports-subpaths-style.ts
```diff
@@ -0,0 +1,149 @@
...
+	return keys.every((key) => !key.startsWith("."));
+}
+
+export const rule = createRule({
+	create(context) {
+		const [{ prefer = "explicit" } = {}] = context.options;
+
+		function validateForExplicit(node: JsonAST.JSONProperty) {
+			if (node.value.type !== "JSONLiteral" && node.value.type !== "JSONObjectExpression") {
+				return;
```
michaelfaith: It looks like there aren't any tests that cover this early return.  Can we add a test that covers this, or Is the condition not necessary?

src/rules/exports-subpaths-style.ts
```diff
@@ -0,0 +1,149 @@
...
+			);
+
+			if (!dotProperty) {
+				return;
+			}
+
+			// Skip unsupported value types
+			if (dotProperty.value.type !== "JSONLiteral" &&
+			    dotProperty.value.type !== "JSONObjectExpression") {
+				return;
```
michaelfaith: It looks like no tests cover these lines.  Can we add coverage for these?

privatenumber:
@michaelfaith 
Thanks for the review—I think this is ready for another review pass.

privatenumber:
My bad, just addressed!

Do we want to add this to the recommended config?

michaelfaith:
> Do we want to add this to the recommended config?

From the conversation [here ](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1322#issuecomment-3415347018)and [here](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1322#issuecomment-3416492076), we thought it would be best to not include it in `recommended`, but possibly consider it for a future `stylistic` config, if there were more rules in that vein in the future.

docs/rules/exports-subpaths-style.md
michaelfaith, line 1: Great work on these docs! 🙌

michaelfaith (approved):
Awesome.  Looks great.  I'll wait a couple days for @JoshuaKGoldberg to weigh in, in case he has any feedback, before merging.  But looks good to me.  Thanks again!

PR #1337, opened 2025-10-23
feat(valid-author): create more precise reports

michaelfaith:
## PR Checklist

## Overview

This change incorporates the new rich output from `package-json-validator`'s `validateAuthor` function, which allows us to report at a more granular level for complex objects.  We're now reporting on the individual properties of Author objects, when violations occur for those properties.

I'm also treating this rule as a prototype for the next generation of factory function for `valid-*` rules, eventually replacing what's in `createSimpleValidPropertyRule`.  In a follow-up change, I'll introduce a new version of that, for the rules powered by the new `package-json-validator` `Result` type.

h/t @christopher-buss for opening the original issue that inspired me to embark on this journey to support more granular reporting.

### DIFF
--- a/src/tests/rules/valid-author.test.ts
+++ b/src/tests/rules/valid-author.test.ts
@@ -9,9 +9,11 @@ ruleTester.run("valid-author", rule, {
 }`,
 			errors: [
 				{
+					data: {
+						error: "the type should be a `string` or an `object` with at least a `name` property",
+					},
 					line: 2,
-					message:
-						"the type should be a `string` or an `object` with at least a `name` property",
+					messageId: "validationError",
 				},
 			],
 			filename: "package.json",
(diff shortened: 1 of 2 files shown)


src/tests/rules/valid-author.test.ts
```diff
@@ -1,150 +1,218 @@
...
-			code: `{ "author": null }`,
+			code: `{
+  "author": null
+}`,
 			errors: [
 				{
+					line: 2,
 					message:
-						'Invalid author: Type for field "author" should be a `string` or an `object` with at least a `name` property',
+						"the type should be a `string` or an `object` with at least a `name` property",
```
JoshuaKGoldberg: [Feature] I liked how it explicitly said `"Invalid author: "` before. Especially if it's a problem with a nested field. Was there a reason you took that out?

michaelfaith: Prior to this change we were batching all violations for each field and listing them all under one report. And in that world, this prefix acted as a sort of header. So, it didn't seem needed once we start reporting them separately at a more granular level. I'm happy to add it back, if you felt it was important, but it'd be different in the sense that it'd be prefixed on every report.  What do you think?

michaelfaith: Added it back

JoshuaKGoldberg (approved):
💯 !

PR #1342, opened 2025-10-25
feat: add new stylistic config

### LINKED ISSUE #1341: 🚀 Feature: add new `stylistic` config
### Feature Request Checklist

### Overview

This conversation came up in the discussion around the new rule @privatenumber recently added: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1322#issuecomment-3415347018

After giving it some more thought, i can already think of some other rules that might be well-suited for a dedicated `stylistic` config.  

- `exports-paths-style` - the rule that triggered this discussion
- `bin-name-casing` (NEW) - we currently have an `enforceCase` option on the `valid-bin` rule that enforces that all names for `bin` entries should be kebab-case.  I'm proposing we introduce a new rule specifically for that (and remove the option from `valid-bin`).
- `script-name-casing` (NEW) - @Zamiell opened [an issue](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/61) to add a similar option to `valid-scripts` that enforces script names should be kebab-case.  Rather than adding it to `valid-scripts`, we could have it be its own rule and include it in this new config.

@JoshuaKGoldberg what do you think? Any other rules you can think of that might be good for this?

### Open Questions
- Logistically, should we add something to the `meta` for rules that we want to include, so that we can automate their inclusion, in the same way we are for `recommended`?  `eslint-plugin-eslint-plugin` uses `meta.docs.category` for this (though, I believe that prop is deprecated...).  We're using it also, but all rules just have "Best Practice" as their category.  Perhaps we can use "Stylistic" for these rules, and have that be the mechanism for including them in the config?

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1341

## Overview

This change adds a new `stylistic` config, which will include rules that are more opinionated about the style of a package.json.  The first rule for this new config is the recently added `exports-subpaths-style`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ import packageJson from "eslint-plugin-package-json";
 
 export default [
 	// your other ESLint configurations
+	packageJson.configs.recommended,
 	packageJson.configs.stylistic,
 ];
 ```


src/plugin.ts
```diff
@@ -76,6 +76,15 @@ const recommendedRules = {
 	],
 } satisfies Linter.RulesRecord;
 
+const stylisticRules = {
+	...Object.fromEntries(
+		Object.entries(rules)
+			// eslint-disable-next-line @typescript-eslint/no-deprecated
```
michaelfaith, line 82: I asked in the ESLint discord if there's a recommended alternative to this prop.  This is the same prop that `eslint-plugin-eslint-plugin` is using for including in its various configs.

README.md
```diff
@@ -61,7 +61,22 @@ export default [
...
+The stylistic configuration enables sets up the parser and files similar to the recommended config, but includes rules that are more opinionated about the style of a package.json.
+This can be used in addition to the recommended config, or on its own.
+
+```ts
+// eslint.config.ts
+import packageJson from "eslint-plugin-package-json";
+
+export default [
+	// your other ESLint configurations
+	packageJson.configs.stylistic,
```
JoshuaKGoldberg, line 76: I think most users of `stylistic` are going to also want `recommended`. How about encouraging them to use both?

```suggestion
	// your other ESLint configurations
	packageJson.configs.recommended,
	packageJson.configs.stylistic,
```

I'm maybe 60% in favor of suggesting this...

PR #1344, opened 2025-10-26
feat(scripts-name-casing): add new rule

### LINKED ISSUE #61: Rule request: kebab-case scripts
It is conventional for scripts in the "package.json" file to be in kebab-case, as evidenced here:

- <https://github.com/angular/angular/blob/main/package.json>
- <https://github.com/facebook/react/blob/main/package.json>
- <https://github.com/microsoft/TypeScript/blob/main/package.json>

Thus, this plugin should warn when incorrectly named scripts exist such as "fooBar" (i.e. camelCase) or "FooBar" (i.e. PascalCase) or "foo_bar" (i.e. snake_case) or "FOO_BAR" (i.e. SCREAMING_SNAKE_CASE).

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #61

## Overview

This change adds a new rule for enforcing that keys of the `scripts` object use `kebab-case`.

This is part of the new `stylelistic` config


docs/rules/scripts-name-casing.md
JoshuaKGoldberg, line 1: [Docs] A lot of repos have scripts like `whatever:this` / `case:is-called`. I think that should be explicitly called out as allowed in the docs & also tested for. It's not clear right now whether that counts as kebab case.

JoshuaKGoldberg (changes requested):
We'd want to include `colon:delineations`, right?

docs/rules/scripts-name-casing.md
michaelfaith: yeah, good callout.  I'll double check that's supported and add tests and docs for it.  And, you know, i don't believe that does have an official name.  Which means we can totally make one 😆.  Colon Kebab Case?  or i guess colon:kebab-case?

michaelfaith: Glad you called that out.  It wasn't supporting it.  But it does now.  Updated!

JoshuaKGoldberg (approved):
Whoop!

PR #1366, opened 2025-11-07
feat(no-redundant-publishConfig): add new rule

### LINKED ISSUE #1365: 🚀 Feature: Bring in a no-redundant-publishConfig rule
### Feature Request Checklist

### Overview

Following #814: we need a `no-redundant-publishConfig` rule that warns when `publishConfig` has no effect.

Per [npm docs](https://docs.npmjs.com/cli/v8/commands/npm-publish), unscoped packages (without `@`) are always public. Using `publishConfig.access` in these packages is redundant since they cannot be made private.

The rule should:

- Warn when unscoped packages have `publishConfig.access` defined
- Remain silent for scoped packages (`@org/name`) where the field is meaningful

❌ Should warn:

```json
{
  "name": "my-package",
  "publishConfig": { "access": "public" }
}
```

✅ Should pass:

```json
{
  "name": "@myorg/my-package",
  "publishConfig": { "access": "public" }
}
```

azat-io:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1365

## Overview

Adds new rule.


src/rules/no-redundant-publishConfig.ts
```diff
@@ -0,0 +1,81 @@
...
+		hasSuggestions: true,
+		messages: {
+			redundantAccess:
+				"'publishConfig.access' is redundant for unscoped packages - they are always public.",
+			removeAccess: "Remove the redundant 'access' field.",
+		},
+		schema: [],
+		type: "suggestion",
+	},
+	name: "no-redundant-publishConfig",
```
azat-io, line 80: I'm a little confused by the camelCase in the rule name. What do you think about renaming it to `no-redundant-publish-config`?

michaelfaith: The property name is `publishConfig`.  so, for all of our rules that are geared around specific properties, we've adopted the convention of using the property name as a single "word" in the rule name.  e.g. `require-publishConfig` = require the `publishConfig` property.  If the rule name were `require-publish-config` that would not be as clear.

michaelfaith: In this case, i could see it both ways.  It could be read as "don't include redundant props within `publishConfig`" in which case the prop name as-is is more precise.  Or it could read as a more generic statement like "don't require unnecessary publish configurations", in which case the prop name wouldn't play in as much. If we wanted to go with the later framing, then I'd prefer to not use the abbreviated word config, and use configuration instead.

michaelfaith:
Hmm, I don't think we've yet discussed whether this should be in `recommended` or not.  Since it's not actively harmful or a source for bugs, should it be?

docs/rules/no-redundant-publishConfig.md
```diff
@@ -0,0 +1,84 @@
...
+{
+	"name": "my-package"
+}
+```
+
+## When Not To Use It
+
+Disable this rule if you:
+
+- Want to keep the `publishConfig.access` field as documentation even though it has no effect
```
michaelfaith, line 83: These are nice 👍

src/rules/no-redundant-publishConfig.ts
```diff
@@ -0,0 +1,81 @@
...
+export const rule = createRule({
+	create(context) {
+		let packageName: string | undefined;
+		let publishConfigAccessProperty: JsonAST.JSONProperty | undefined;
+
+		return {
+			"Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.value=name]"(
+				node: JsonAST.JSONProperty,
+			) {
+				if (isJSONStringLiteral(node.value)) {
```
michaelfaith, line 18: It looks like no tests cover the false result of this condition.  Can a new test be added for that?

src/tests/rules/no-redundant-publishConfig.test.ts
```diff
@@ -0,0 +1,136 @@
...
+						{
+							messageId: "removeAccess",
+							output: `{
+	"name": "my-package",
+	"publishConfig": {
+		
+		"registry": "https://example.com"
+	}
+}`,
+						},
```
michaelfaith, line 70: Would you mind adding `line` props for these errors?  We're trying to be better about the precision of our reports, and adding lines for tests helps us ensure we're doing well with that.

michaelfaith (changes requested):
Thanks for the contribution!  I had a couple feedback items, and then the question about whether or not this should be in the `recommended` config.

src/rules/no-redundant-publishConfig.ts
azat-io: Done ✅️️️️

michaelfaith:
You've convinced me 👍

michaelfaith (approved):
Looks good!  Thanks again 🥇

PR #1393, opened 2025-11-14
feat(valid-repository): add new rule for validating `repository`

### LINKED ISSUE #838: 🚀 Feature: Bring in a valid-repository rule
### Bug Report Checklist

### Overview

Following #42 -> #51: we've resolved to create a standalone `valid-repository` rule. It should enforce that, _if_ the field exists in the `package.json` file, it's set to a valid value. It should do nothing if the field doesn't exist.

### Additional Info

Per discussion in #51, this won't be enabled by default at first. The `valid-package-def` rule already contains this logic using [`package-json-validator`](https://github.com/JoshuaKGoldberg/package-json-validator). For this issue, we should:

💖

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #838

## Overview

This change adds a new `valid-repository` rule that uses `validateRepository` from package-json-validator to identify errors.  It's included in the `recommended` config.


src/tests/rules/valid-repository.test.ts
```diff
@@ -0,0 +1,211 @@
...
+			"bitbucket:user/repo",
+			"gitlab:user/repo",
+		].map((value) => ({
+			code: `{
+	"repository": "${value}"
+}
+`,
+			name: value,
+		})),
+	],
```
JoshuaKGoldberg, line 210: I'm not a huge fan of auto-generating tests. I've advocated for banning them altogether in typescript-eslint, and will eventually try to spread the practice elsewhere: https://github.com/typescript-eslint/typescript-eslint/issues/11052. But not a blocker IMO. This repo is a lot simpler than typescript-eslint's.

michaelfaith: Hmm, I'm not sure I fully agree in this case.  I think, like with anything, it can be taken too far, and result in hard to reason about tests.  For instance, I don't at all like the way many of the test suites in `eslint-plugin-import` are constructed (e.g. https://github.com/import-js/eslint-plugin-import/blob/main/tests/src/rules/no-cycle.js#L287-L311).  I've always found those harder to work with and a bit unapproachable.  

But in simple cases like this where the test code is all going to be exactly the same except for one prop value, it's an easy way to reduce some boilerplate.  I guess what I'm trying to say is that it doesn't feel like quite so black and white to me, and extremes are bad for most things 😄

PR #1398, opened 2025-11-14
feat(unique-dependencies): report cross-group duplicates

### LINKED ISSUE #1007: 🐛 Bug: `unique-dependencies` doesn't flag duplicates between dependencies and devDependencies

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1007

## Overview

This change adds the ability to report duplicate dependencies across dependency groups.  Specifically, if dependencies appear in `peer-` or `dev-` dependencies that are already in `dependencies`, those dependencies are also reported.

### DIFF
--- a/src/rules/unique-dependencies.ts
+++ b/src/rules/unique-dependencies.ts
@@ -130,7 +130,6 @@ export const rule = createRule({
 							isJSONStringLiteral(otherDependencyKey) &&
 							dependencyNames.has(otherDependencyKey.value)
 						) {
-							// We have a duplicate, report it
 							context.report({
 								messageId: "crossGroupDuplicate",
 								node: otherDependencyNode,


src/rules/unique-dependencies.ts
```diff
@@ -88,9 +95,59 @@ export const rule = createRule({
...
+					"peerDependencies",
+				]) {
+					const otherDependencies = dependenciesCache[dependencyType];
+					for (const otherDependencyNode of otherDependencies) {
+						const otherDependencyKey = otherDependencyNode.key;
+						if (
+							isJSONStringLiteral(otherDependencyKey) &&
+							dependencyNames.has(otherDependencyKey.value)
+						) {
+							// We have a duplicate, report it
```
JoshuaKGoldberg: Nit: I think this is implied by the fact that it's a `context.report`?

```suggestion
```

JoshuaKGoldberg (approved):
Awesome!!

PR #1408, opened 2025-11-16
feat(require-attribution): add new rule to require proper attribution

### LINKED ISSUE #1404: 🚀 Feature: create new rule that requires attribution
### Feature Request Checklist

### Overview

As resolved in #1378, we'd like to provide a new `recommended-` config specifically geared towards publishable packages.  As part of this new config, we'd like a rule that requires that proper attribution is included in the `package.json`.  The rule should ensure that either `author` or `contributors` is included, and that if `contributors` is present, it should include at lease one contributor.  

The rule should also have an option that enforces that only one of the props is present and not both (should this be true by default?).

### Additional Info

_No response_

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1404

## Overview

This change adds a new `require-attribution` rule, which requires that either `author` or `contributors` is defined, and that if `contributors` is present, it should include at least one contributor.

### DIFF
--- a/src/rules/require-attribution.ts
+++ b/src/rules/require-attribution.ts
@@ -27,8 +27,7 @@ export const rule = createRule({
 				const contributorsValue = node.value;
 				if (
 					contributorsValue.type !== "JSONArrayExpression" ||
-					contributorsValue.elements.filter((element) => !!element)
-						.length === 0
+					!contributorsValue.elements.some((element) => !!element)
 				) {
 					context.report({
 						messageId: "noContributors",


src/utils/createSimpleRequirePropertyRule.ts
```diff
@@ -87,6 +87,8 @@ export const createSimpleRequirePropertyRule = (
 					properties: {
 						ignorePrivate: {
 							default: ignorePrivateDefault,
+							description:
+								"Determines if this rule should be enforced when the package's `private` property is `true`.",
```
JoshuaKGoldberg, line 91: +1

src/rules/require-attribution.ts
```diff
@@ -0,0 +1,107 @@
...
+				node: JsonAST.JSONProperty,
+			) {
+				contributorsPropertyNode = node;
+
+				// Ensure that there's at least one contributor.
+				const contributorsValue = node.value;
+				if (
+					contributorsValue.type !== "JSONArrayExpression" ||
+					contributorsValue.elements.filter((element) => !!element)
+						.length === 0
```
JoshuaKGoldberg: Nit: less code, and also a general way to exit loops early (not that this perf would ever matter)

```suggestion
					!contributorsValue.elements.some((element) => !!element)
```

JoshuaKGoldberg (approved):
📛

PR #1424, opened 2025-11-29
chore: configure eslint-config-un for linting

andreww2012:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1423
- [ ] That issue was marked as [`status: accepting prs`](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

This PR migrates this repository to my ESLint config generator, [eslint-config-un](https://github.com/andreww2012/eslint-config-un/tree/dev-v1). The commits follow [the config migration guide](https://github.com/andreww2012/eslint-config-un/tree/dev-v1?tab=readme-ov-file#migrating-existing-codebase-to-eslint-config-un). Refer to #1423 for more info.

While many stylistic issues have been fixed, a few actual problems have been found.

Issue #1332, opened 2025-10-19
🛠 Tooling: set up trusted publisher


michaelfaith:
### Tooling Report Checklist

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have pulled the latest `main` branch of the repository.
- [x] I have [searched for related issues](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues?q=is%3Aissue) and found none that matched my issue.

### Overview

To improve security, i'd propose we shift to npm's trusted publisher method for package publishing, and removing the long-lived token.

https://docs.npmjs.com/trusted-publishers

JoshuaKGoldberg:
Agreed, yes, we should. I'm a little swamped this month but definitely want to get it into CTA and downstream packages. https://github.com/JoshuaKGoldberg/create-typescript-app/issues/2270

PR #1429, opened 2025-12-01
chore: eslint config improvements

### LINKED ISSUE #1423: 🛠 Tooling: use eslint-config-un for linting
### Tooling Report Checklist

### Overview

Hello! For more than a year, I've been working extensively on [eslint-config-un](https://github.com/andreww2012/eslint-config-un/tree/dev-v1), a comprehensive ESLint config generator that grew out of the personal collection of rules. I believe it is now mature enough to propose as a real option for an open-source project to adopt.

Main features are:
- It's opinionated, but has been tested across numerous public repositories and refined to ensure sensible defaults for most use cases; at the same time it's *extremely* configurable;
- It includes almost 100 ESLint plugins - from the most popular to niche ones, while keeping most of them as optional peer dependencies;
- Everything is strictly typed (types for all known rules' options are pre-generated with `eslint-typegen`);
- It allows bringing your own plugins, overriding plugin implementations, and changing plugin prefixes with ease;
- Has much more unique features that make it suitable for as much as for projects and scenarios - it's all covered [in the config documentation](https://github.com/andreww2012/eslint-config-un/tree/dev-v1#readme).

***Fun fact**: [my contributions to eslint-plugin-package-json](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pulls?q=is%3Apr+author%3Aandreww2012+is%3Aclosed) was made to address the issues I've found during config testing.*

Since I've contributed to this project before and it is of moderate size, I think it's the perfect candidate for real world config testing. I've already put up [the draft PR](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1424) which followed [the migration guide](https://github.com/andreww2012/eslint-config-un/tree/dev-v1?tab=readme-ov-file#migrating-existing-codebase-to-eslint-config-un) and has my comments better explaining the changes and made decisions.

I would be happy to receive any feedback on the project and answer any questions. It would also be understandable if you decide not to proceed with this. Thank you!

### Additional Info

_No response_

andreww2012:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1423

## Overview

This PR was based on issues found in https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1424 and includes the following:

- Added `eslint-plugin-markdown-links` for checking local & remote links and fixed found dead URLs;
- ~Enabled ESLint caching to slightly speed up local development;~
- Enabled `@typescript-eslint/no-shadow` and fixed found issues;
- ~Added `eslint-plugin-depend` and replaced `lint-staged` with `nano-staged` as per its recommendation;~
- Added `eslint-plugin-node-dependencies` (no issues found);
- Added `eslint-plugin-security` (IMHO too strictly setup by default, so I turned off some rules and suppressed some false positives on regexes);
- Added `eslint-plugin-unicorn`'s opinionated config and mostly suppressed some rules;
- ~Added `eslint-plugin-sonarjs`, same deal as with `eslint-plugin-unicorn`.~

### DIFF
--- a/src/rules/sort-collections.ts
+++ b/src/rules/sort-collections.ts
@@ -58,7 +58,6 @@ export const rule = createRule({
 				const currentOrder = collection.properties;
 				let desiredOrder: JsonAST.JSONProperty[];
 
-				// If it's any property other than `scripts`, simply sort lexicographically
 				if (keyPartsReversed.at(-1) === "scripts") {
 					// For scripts we'll use `sort-package-json`
 					const scriptsSource = context.sourceCode.getText(node);
@@ -79,6 +78,8 @@ export const rule = createRule({
 					desiredOrder = Object.keys(sortedScripts).map(
 						(prop) => propertyNodeMap[prop],
 					);
+
+					// If it's any property other than `scripts`, simply sort lexicographically
 				} else {
 					desiredOrder = currentOrder.toSorted((a, b) => {
 						const aKey = (a.key as JsonAST.JSONStringLiteral).value;


JoshuaKGoldberg:
> Enabled ESLint caching to slightly speed up local development

https://typescript-eslint.io/troubleshooting/faqs/eslint/#can-i-use-eslints---cache-with-typescript-eslint 🥲

.husky/pre-commit
```diff
@@ -1 +1 @@
-npx lint-staged
+npx nano-staged
```
michaelfaith: Are we sure this change is the direction we want to go?  Just looking at adoption and project stats: 50k weekly downloads vs 14 million, and it's not reached 1.0 yet?  I like the work the e18e folks are doing in evangelizing package modernization, but they can be a bit too aggressive sometimes imho.  I'm not sure that a lint plugin enforcing their recommendations is the way to go.

src/tests/index.test.ts
```diff
@@ -1,6 +1,6 @@
 import { ESLint, Linter } from "eslint";
 import { readFile } from "node:fs/promises";
-import { resolve } from "node:path";
+import path from "node:path";
```
michaelfaith, line 3: Curious about this change.  Why move away from a named import here?

src/plugin.ts
```diff
@@ -84,6 +79,11 @@ const stylisticRules = {
 	),
 } satisfies Linter.RulesRecord;
 
+const { name, version } = require("../package.json") as {
+	name: string;
+	version: string;
+};
+
```
michaelfaith, line 84: What's the need for this change?

eslint.config.ts
```diff
@@ -2,14 +2,23 @@
...
+import depend from "eslint-plugin-depend";
 import eslintPlugin from "eslint-plugin-eslint-plugin";
 import jsdoc from "eslint-plugin-jsdoc";
 import jsonc from "eslint-plugin-jsonc";
+import markdownLinks from "eslint-plugin-markdown-links";
 import n from "eslint-plugin-n";
+import nodeDependencies from "eslint-plugin-node-dependencies";
 import perfectionist from "eslint-plugin-perfectionist";
+// @ts-expect-error - no types
+import promise from "eslint-plugin-promise";
```
michaelfaith: Do we really need this one?  We don't tend to work with promises directly much in this repo.

eslint.config.ts
```diff
@@ -2,14 +2,23 @@
 import comments from "@eslint-community/eslint-plugin-eslint-comments/configs";
 import eslint from "@eslint/js";
 import vitest from "@vitest/eslint-plugin";
+import depend from "eslint-plugin-depend";
```
michaelfaith: I raised this in another comment, but not sure we should include this. I'm good with adopting newer packages strategically, but having that be a forced activity via lint plugin, I'm less enthusiastic about.

eslint.config.ts
```diff
@@ -133,4 +157,40 @@ export default defineConfig(
...
+						String.raw`/^https:\/\/(?:www\.)?npmjs.com\/.*/`,
+					],
+					maxRetries: 3,
+					timeout: 5000,
+				},
+			],
+		},
+	},
+	{
+		extends: [...nodeDependencies.configs["flat/recommended"]],
```
michaelfaith: I'm not sure we get much from this config.  There are only three rules enabled in the config, and two of them we already have within our plugin.  I wouldn't want to have multiple rules for the same thing.  It seems like only the compat-engines rule in the config is one that we don't have coverage for already.  Should we just use that one rule, rather than the whole config?  I'm also kind of ok with just dropping that entirely anyway, since `pnpm install` will fail if the engines aren't compatible, before any eslint rule violation fires.  I'd advocate for just dropping the the config entirely.

eslint.config.ts
```diff
@@ -133,4 +157,40 @@ export default defineConfig(
...
+		files: ["**/*.md"],
+		ignores: ["CHANGELOG.md"],
+		rules: {
+			"markdown-links/no-dead-urls": [
+				"error",
+				{
+					checkAnchor: false,
+					ignoreUrls: [
+						// npm gets rate-limited quickly: https://github.com/ota-meshi/eslint-plugin-markdown-links/issues/42
+						String.raw`/^https:\/\/(?:www\.)?npmjs.com\/.*/`,
```
michaelfaith: Just thinking about the process of creating a brand new rule.  If someone's creating a new rule that links to a new doc page that hasn't been merged yet, are they going to have to fight this rule to get that merged?

eslint.config.ts
```diff
@@ -2,14 +2,23 @@
...
 import jsonc from "eslint-plugin-jsonc";
+import markdownLinks from "eslint-plugin-markdown-links";
 import n from "eslint-plugin-n";
+import nodeDependencies from "eslint-plugin-node-dependencies";
 import perfectionist from "eslint-plugin-perfectionist";
+// @ts-expect-error - no types
+import promise from "eslint-plugin-promise";
 import * as regexp from "eslint-plugin-regexp";
+import security from "eslint-plugin-security";
+import sonarjs from "eslint-plugin-sonarjs";
```
michaelfaith: I question whether we should use this plugin.  It seems to be duplicating a number of rules from other plugins and is a sort of grab bag of several different kinds of issues, rather than having a central theme, like other plugins.  It's main purpose seems to be to align with the sonarqube platform, which we're not using. It also has an LGPL-3.0-only license, which is not super permissive.  Kind of prefer removing this.

src/rules/unique-dependencies.ts
```diff
@@ -37,18 +37,16 @@ export const rule = createRule({
 			for (const element of elements
 				.filter(isNotNullish)
 				.filter(isJSONStringLiteral)
+				// eslint-disable-next-line unicorn/no-array-reverse
```
michaelfaith: Should we address instead of disable?

src/rules/restrict-dependency-ranges.ts
```diff
@@ -76,13 +76,15 @@ const changeVersionRange = (version: string, rangeType: RangeType): string => {
...
 				return "workspace:*";
+			// eslint-disable-next-line unicorn/no-useless-switch-case
 			case "tilde":
 			default:
 				return "workspace:~";
 		}
 	}
 
 	return version.replace(
+		// eslint-disable-next-line security/detect-unsafe-regex -- false positive
```
michaelfaith: Considering the number of false positives, should we turn this rule off entirely?  Doesn't seem very reliable.

src/rules/restrict-dependency-ranges.ts
```diff
@@ -76,13 +76,15 @@ const changeVersionRange = (version: string, rangeType: RangeType): string => {
 				return "workspace:^";
 			case "pin":
 				return "workspace:*";
+			// eslint-disable-next-line unicorn/no-useless-switch-case
```
michaelfaith: Should we address instead of disable?

src/rules/sort-collections.ts
```diff
@@ -50,6 +50,7 @@ export const rule = createRule({
 						return;
 					}
 				}
+				// eslint-disable-next-line unicorn/no-array-reverse, sonarjs/no-misleading-array-reverse
```
michaelfaith: Should we address instead of disable?

.husky/pre-commit
andreww2012: I agree that some e18e efforts can be rather radical, but in this case this replacement seemed safe to me. The pre-commit hook continued to work as before, and the size of this repo is not large enough to make switching dev tools a big concern. As for the possibility of other potentially unwanted replacement suggestions, it's possible [to suppress them](https://github.com/es-tooling/eslint-plugin-depend/blob/HEAD/docs/rules/ban-dependencies.md#allowed).

Anyway this is just my opinion, and the final decision is yours - I'll revert this change if needed.

src/tests/index.test.ts
andreww2012: This comes from [`unicorn/import-style`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v62.0.0/docs/rules/import-style.md). While I agree with this particular change because using default import slightly reduces mental load (it has become an unwritten rule to name `path` module import `path`, making it immediately clear where `path` comes from, as opposed to just `resolve`), I don't understand why this rule ended up in the unopinionated config which I have enabled in my PR.

src/plugin.ts
andreww2012: I enabled [@typescript-eslint/no-shadow](https://typescript-eslint.io/rules/no-shadow) so if I had not moved this `name` definition closer to its usage (or renamed the variable, which is another valid option), the rule would report an error in the `.map(([name])` areas. I decided to go with this option because bringing the variable definition as close as possible to its usage generally improves readability.

eslint.config.ts
andreww2012: Can revert this one, but it costed almost zero effort to set up. I think it won't harm to keep this plugin enabled, and it might become useful if one day you start work promises more here.

eslint.config.ts
andreww2012: Replied in that comment - unwanted suggestions can be suppressed.

eslint.config.ts
andreww2012: I agree that the recommended config might not be very useful, but I've also enabled not currently implemented in `eslint-plugin-package-json` [`no-deprecated`](https://ota-meshi.github.io/eslint-plugin-node-dependencies/rules/no-deprecated.html) rule. Maybe I should remove the recommended config and leave that one rule enabled?

Ultimately, your plugin and this one serve a very similar purpose, so the ideal outcome would be to have all the rules consolidated into a single plugin ☺️

eslint.config.ts
andreww2012: Good point - adding rule docs pages to `ignoreUrls`.

eslint.config.ts
andreww2012: Removed it.

src/rules/restrict-dependency-ranges.ts
andreww2012: Since `eslint-plugin-regexp` is also enabled, I agree that this rule doesn't add much value. Disabled

src/rules/restrict-dependency-ranges.ts
andreww2012: Usually switch exhaustiveness is a good thing - so I'd prefer to leave this `case` (and also not to duplicate `return "workspace:~"`). Since we currently do not enforce exhaustiveness, it would be logical to enable [`@typescript-eslint/switch-exhaustiveness-check`](https://typescript-eslint.io/rules/switch-exhaustiveness-check) that would reduce the incentive to remove this "useless switch case". What do we do in the end?

src/rules/sort-collections.ts
andreww2012: In all the suppressed `unicorn/no-array-reverse` reports the usage of `reverse` looked safe to me - I still think this rule is useful, but overly strict.

src/rules/unique-dependencies.ts
andreww2012: Replied in another comment

eslint.config.ts
michaelfaith: > Maybe I should remove the recommended config and leave that one rule enabled?

That sounds good to me 👍

eslint.config.ts
michaelfaith: > Can revert this one

Yes please

.husky/pre-commit
michaelfaith: Yes please

src/rules/restrict-dependency-ranges.ts
michaelfaith: > Usually switch exhaustiveness is a good thing

Agreed.  Then should we just disable this rule?  Anytime I encounter rules where I find myself reaching for inline disables on situations that are actually legitimate (and not just tech debt that needs to be addressed), it makes me question the whether the rule is actually value-add, or just noise.  I would say, let's disable it.

src/rules/sort-collections.ts
michaelfaith: In that case, I would suggest we disable it.  For similar reasons as https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1429#discussion_r2588664874

eslint.config.ts
michaelfaith: Even so, that doesn't change the fundamental concern of not wanting to use an eslint rule to force dependency decisions.  I'd rather remove the plugin.

src/rules/restrict-dependency-ranges.ts
andreww2012: Enabled `@typescript-eslint/switch-exhaustiveness-check` and disabled `unicorn/no-useless-switch-case`. Had to transform 2-cases switch to `if-else` block to avoid suppressing `@typescript-eslint/switch-exhaustiveness-check`.

eslint.config.ts
```diff
@@ -41,9 +46,13 @@ export default defineConfig(
 			n.configs["flat/recommended"],
 			perfectionist.configs["recommended-natural"],
 			regexp.configs["flat/recommended"],
+			security.configs.recommended,
+			unicorn.configs.unopinionated,
 		],
 		files: JS_TS_FILES,
 		rules: {
+			"@typescript-eslint/no-shadow": "error",
+			"@typescript-eslint/switch-exhaustiveness-check": "error",
```
michaelfaith: Do we actually want this rule on though? I know you mentioned it in the conversation around unicorn/no-useless-switch-case, but it feels overly strict in the other direction. Particularly with ESTree / JSONAst types, which have string unions with so many options. It's not uncommon to use switch statements that just pick out a handful of the options. The case you changed only had two, so it wasn't a big deal to move that to if conditions. But if it had like 5 or 6, then it would have been a bit more of a mess. I'd rather leave this off, too.

eslint.config.ts
```diff
@@ -133,4 +148,41 @@ export default defineConfig(
...
+			// @ts-expect-error types mismatch
+			"node-dependencies": nodeDependencies,
+		},
+		rules: {
+			"node-dependencies/no-deprecated": [
+				"error",
+				{ devDependencies: true },
+			],
+		},
+	},
```
michaelfaith: Since this targets the same file as the config on line 99, can they be combined?

michaelfaith (changes requested):
Good progress!  Just a couple more items on my side.

eslint.config.ts
andreww2012: I agree that this rule can be a double-edged sword sometimes - so I guess the decision to keep it enabled comes down to whether one is ready to put up with the extra `eslint-disable` comments. Removed this rule.

.github/CONTRIBUTING.md
```diff
@@ -87,7 +87,7 @@ Once all feedback is addressed and the PR is approved, we'll ensure the branch i
 
 #### Post-Merge Recognition
 
-Once your PR is merged, if you haven't yet been added to the [_Contributors_ table in the README.md](../README.md#contributors) for its [type of contribution](https://allcontributors.org/docs/en/emoji-key "Allcontributors emoji key"), you should be soon.
+Once your PR is merged, if you haven't yet been added to the [_Contributors_ table in the README.md](../README.md#contributors) for its [type of contribution](https://allcontributors.org/emoji-key "Allcontributors emoji key"), you should be soon.
```
JoshuaKGoldberg: Oh, interesting, I didn't know this was one of the 404s. This is IMO definitely a bug in the All Contributors side. Old links like this really shouldn't 404. https://github.com/all-contributors/all-contributors.github.io/issues/979

(not a request for changes, +1 to fixing to not 404 in the interim)

src/rules/order-properties.ts
```diff
@@ -58,10 +58,10 @@ export const rule = createRule({
...
 
 				const allKeys = Object.keys(json);
-				const nonStandardKeys = allKeys.filter(
-					(key) => !requiredOrder.includes(key),
-				);
-				const orderedNonStandardKeys = nonStandardKeys.sort();
+				const orderedNonStandardKeys = allKeys
+					.filter((key) => !requiredOrder.includes(key))
+					// eslint-disable-next-line unicorn/no-array-sort
+					.sort();
```
JoshuaKGoldberg, line 63: -1 from me on enabling this rule pending https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2735#issuecomment-3637969180. Out of the common benefits of the rule:

* Performance: we don't deal with large enough sets of data for that to be a bottleneck
* Testing: we already have relatively high test coverage, so it's not likely to catch many bugs

(obviously linting is nice as an earlier line of defense - I'm saying the benefits are severely reduced, not eliminated)

src/rules/order-properties.ts
```diff
@@ -97,7 +97,7 @@ export const rule = createRule({
 									type === "tab" ? "\t" : indent,
 								) + endCharacters;
 							if (newline === "\r\n") {
-								result = result.replace(/\n/g, newline);
+								result = result.replaceAll("\n", newline);
```
JoshuaKGoldberg: Another -1 from me on the specific rule. The rule is asking for a stylistic preference that actually makes the code more verbose.

src/rules/restrict-dependency-ranges.ts
```diff
@@ -76,7 +76,6 @@ const changeVersionRange = (version: string, rangeType: RangeType): string => {
 				return "workspace:^";
 			case "pin":
 				return "workspace:*";
-			case "tilde":
```
JoshuaKGoldberg, line 79: [Praise] I like this 🙂

src/rules/restrict-dependency-ranges.ts
```diff
@@ -112,7 +111,7 @@ export const rule = createRule({
 
 		// Reverse the array, so that subsequent options override previous ones
 		const optionsProvided = Array.isArray(context.options[0])
-			? [...context.options[0]].reverse()
+			? context.options[0].toReversed()
```
JoshuaKGoldberg, line 114: [Praise] I like this 🙂

src/tests/rules/order-properties.test.ts
```diff
@@ -440,6 +440,26 @@ ruleTester.run("order-properties", rule, {
...
+					data: { property: "name" },
+					messageId: "incorrectOrder",
+				},
+			],
+			filename: "package.json",
+			options: [{ order: ["name", "version"] }],
+			output: ["{", '	"name": "foo",', '	"version": "1.0.0"', "}"].join(
+				"\r\n",
+			),
+		},
```
JoshuaKGoldberg, line 462: Can you please split out any real behavior changes + tests into their own PRs? It makes it hard to review tooling ones like this.

OTOH, if this is just to appease Codecov, I'd say not to worry about it. Codecov falsely reporting a coverage change is just a bug in the concept of test coverage. I'd rather just have the build failure than do unnecessary work.

src/utils/isPackageJson.test.ts
```diff
@@ -19,11 +19,11 @@ describe("isPackageJson", () => {
...
 		["/package.json", true],
-		["\\package.json", true],
+		[String.raw`\package.json`, true],
 		["prefix/package.json", true],
-		["prefix\\package.json", true],
-		["mixed\\prefix/package.json", true],
-		["mixed/prefix\\package.json", true],
+		[String.raw`prefix\package.json`, true],
+		[String.raw`mixed\prefix/package.json`, true],
+		[String.raw`mixed/prefix\package.json`, true],
```
JoshuaKGoldberg: -1 from me on this stylistic preference that causes more code.

src/plugin.ts
JoshuaKGoldberg: -1 from me, I don't think that rule's useful. It's not enabled in the tseslint recommended configs.

Edit: eh, on second thought, I do kind of like the removal of duplicate names... 👍

eslint.config.ts
```diff
@@ -5,9 +5,13 @@ import vitest from "@vitest/eslint-plugin";
 import eslintPlugin from "eslint-plugin-eslint-plugin";
 import jsdoc from "eslint-plugin-jsdoc";
 import jsonc from "eslint-plugin-jsonc";
+import markdownLinks from "eslint-plugin-markdown-links";
 import n from "eslint-plugin-n";
+import nodeDependencies from "eslint-plugin-node-dependencies";
 import perfectionist from "eslint-plugin-perfectionist";
 import * as regexp from "eslint-plugin-regexp";
+import security from "eslint-plugin-security";
```
JoshuaKGoldberg: I've never seen a codebase actually benefit from `eslint-plugin-security` (when `eslint-plugin-regexp` is in play).

* The security plugin is much more applicable for programs that run on user input, receive requests, etc. - not so much walled-garden lint plugins
* It's not very actively maintained

-1 from me on including it

eslint.config.ts
```diff
@@ -82,6 +94,16 @@ export default defineConfig(
 	{
 		extends: [packageJson.configs["recommended-publishable"]],
 		files: ["package.json"],
+		plugins: {
+			// @ts-expect-error types mismatch
```
JoshuaKGoldberg: Can you link to a tracking issue? (if one doesn't exist, let's file one)

JoshuaKGoldberg (changes requested):
Thanks for proactively working on this, and sorry for the delay in reviewing! There are some parts I quite like and think are going to be nice to have. And some parts I'm not so in favor of. What do you think?

(re-requesting michaelfaith's review to loop into threads explicitly)

eslint.config.ts
michaelfaith: I've never used it myself, and didn't have an opinion one way or another.  Totally agree with your points, and especially if it's not well-maintained.

src/tests/rules/order-properties.test.ts
michaelfaith: This was my ask.  Codecov was complaining, but correctly so.  We didn't have coverage for non-linux/mac use cases of this rule, yet we have logic to support windows.  With that said, we have unit tests running on windows, we're just not collecting coverage for the windows test runs.  So it wasn't a bug so much as a blind spot.  I'm also ok if this is broken out, but codecov got triggered because this change touched a line in the windows-only logic.

src/rules/order-properties.ts
michaelfaith: I could go either way.  Definitely more of a stylistic thing, but in this case I kind of like the suggestion (less so on the String.raw change 😬). So, I'm 🤷

src/utils/isPackageJson.test.ts
michaelfaith: Agreed

src/rules/order-properties.ts
michaelfaith: Gotta love the insta-close after your comment 😭 .  In this case, we should probably turn the rule off, as they suggested?

andreww2012: It's a shame that they decided to not make this rule smarter, yet it really requires type information to avoid false positives and false negatives. Disabled this rule.

src/rules/order-properties.ts
andreww2012: It is once again interesting that [`unicorn/prefer-string-replace-all` is in the `unopinionated` config](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/HEAD/docs/rules/prefer-string-replace-all.md) 🤷‍♂️

Disabled this rule, but IMHO the slight verbosity that this rule suggests make the code a bit easier to read.

src/utils/isPackageJson.test.ts
andreww2012: Yet another [`unopinionated` unicorn rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/HEAD/docs/rules/prefer-string-raw.md) 🤷‍♂️

I personally do like it because I don't mind adding a bit of verbosity to improve readability. Disabled this rule.

eslint.config.ts
andreww2012: Decided to add it just and case and actually agree with these points. Removed it

eslint.config.ts
andreww2012: Done

src/rules/order-properties.ts
JoshuaKGoldberg: Yeah I'm starting to regret including the rule in it... I bet this is one of a few _"let's move out of `unopinionated`"_ issues we should file.

src/tests/rules/order-properties.test.ts
JoshuaKGoldberg: Gotcha gotcha. I'm in favor of moving out, yeah, but not vehemently so.

.github/CONTRIBUTING.md
JoshuaKGoldberg: All Contributors is fixed now! https://allcontributors.org/docs/en/emoji-key works again. This change should be reverted. 🥳

andreww2012: Done

src/rules/sort-collections.ts
```diff
@@ -59,14 +59,7 @@ export const rule = createRule({
 				let desiredOrder: JsonAST.JSONProperty[];
 
 				// If it's any property other than `scripts`, simply sort lexicographically
```
michaelfaith: Nit: it'd be nice to move this comment to the condition it applied to (now the `else`)

andreww2012: Agree 👍

eslint.config.ts
```diff
@@ -62,6 +67,10 @@ export default defineConfig(
 			"no-useless-rename": "error",
 			"object-shorthand": "error",
 			"operator-assignment": "error",
+			"unicorn/no-array-reverse": "off", // Overly strict
+			"unicorn/no-array-sort": "off", // Overly strict
+			"unicorn/prefer-string-raw": "off",
+			"unicorn/prefer-string-replace-all": "off",
```
JoshuaKGoldberg: [Style] Nit: to not have to write the same thing multiple times, and aligning with previous "groups of rules" comments:

```suggestion
			// These on-by-default rules are too strict for our preferences
			"unicorn/no-array-reverse": "off",
			"unicorn/no-array-sort": "off",
			"unicorn/prefer-string-raw": "off",
			"unicorn/prefer-string-replace-all": "off",
```

michaelfaith:
> I suppose we could merge this before https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/1332 is resolved and just not publish to users.

I guess we could.  It's just that every merge is going to result in a failed Release build, which is kind of 😭 .  But you're right, we could technically still merge, and whenever the next change is merged after its fixed, it'll pick all the changes up in that release.

PR #1480, opened 2025-12-31
feat: add require-homepage rule

### LINKED ISSUE #865: 🚀 Feature: Bring in a require-homepage rule
### Bug Report Checklist

### Overview

Following #42 -> #51: we've resolved to create a standalone `require-homepage` rule. It should enforce that the field exists in the `package.json` file.

### Additional Info

💖

marcalexiei:
## PR Checklist

- [x] Addresses an existing open issue: fixes #865

## Overview

Add `homepage` to properties listed in `src/rules/require-properties.ts`


docs/rules/require-homepage.md
```diff
@@ -0,0 +1,78 @@
+# require-homepage
+
+💼 This rule is enabled in the 📦 `recommended-publishable` config.
```
michaelfaith: Interesting.  We didn't really discuss this being part of the publishable config.  ~Seems reasonable but just want to double check, should it be?  What do you think JoshuaKGoldberg?  (We'll need to mark this as breaking, if we include this change. - which is totally cool, but highlighting it as a change needed)~

src/utils/createSimpleRequirePropertyRule.ts
```diff
@@ -7,7 +7,7 @@ export interface CreateRequirePropertyRuleOptions {
 	/**
 	 * The category for this rule.
 	 */
-	category?: string;
```
michaelfaith, line 10: I think we should leave this as `string`, since there could be a world (though, admittedly, unlikely) that we include one of these rules in some other config like Stylelistic.

docs/rules/require-homepage.md
michaelfaith: Oh, actually, I took a second to think about this, and _don't_ think this should be in the publishable config.  There are different kinds of publishing that happen.  Some are public, which is where this would be good to require, but every company I've worked for in the last 15 years has also had an internal registry where packages can be published for internal use.  And it wouldn't make sense to require `homepage` in those cases.  So, let's not include it for now.

marcalexiei: > So, let's not include it for now.

Gotcha

---

>  but every company I've worked for in the last 15 years has also had an internal registry where packages can be published for internal use. 

For this reason maybe we should set `ignorePrivateDefault` to `true`?

src/utils/createSimpleRequirePropertyRule.ts
marcalexiei: I used a template literal instead of a plain `string` because `Published` is already used in this codebase (see reference below).

https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/blob/6d83a7f0b815cb1784761428950f01e6f50aa2ee/src/plugin.ts#L66

Since some rules may eventually live in the `Stylistic` config as well, modeling this as a union (`'Published' | 'Stylistic'`) might be more appropriate.

If fully arbitrary values are required, we can extend the union with `string & {}`. 
This retains IDE autocomplete for the known values while still allowing flexibility and reducing the risk of typos.

docs/rules/require-homepage.md
michaelfaith: > For this reason maybe we should set ignorePrivateDefault to true?

That wouldn't really work.  `private` in a package.json is different from an internally hosted corporate registry.  You still have to have `private: true` in order to publish to _any_ registry, whether it's npmjs.org or an internal registry.

src/utils/createSimpleRequirePropertyRule.ts
michaelfaith: The second of those options would be my preference, just so we don't have yet another place that would need to be updated if we created another new config in the future.

michaelfaith (approved):
Looks good! Thanks again 👍

michaelfaith:
@all-contributors please add @marcalexiei for code

Issue #850, opened 2025-02-01
💬 Discussion: Which `require-` and `valid-` rules should be in the recommended config?


michaelfaith:
Now that we've landed on a long-term strategy for the `require-` and `valid-` flavor of rules.  I'd like to start the discussion of which `require-` and  `valid-` rules should we include in the `recommended` config?

I'll start a table here with all that have issues created, and we can use this thread to discuss which of them should belong in the config and which shouldn't.

## Require

| rule | issue | recommended |
|------|-------|-------------|
|  `require-author`    | #795       |             |
| `require-bin`     | #796      |             |
| `require-bugs`     | #862|             |
|  `require-bundleDependencies`    | #797      |             |
| `require-config`     | #798|             |
| `require-contributors`     | #799|             |
| `require-cpu`     | #800|             |
| `require-dependencies`     | #863|             |
| `require-description` |  #864 | ✅ |
| `require-devDependencies`     | #801      |             |
| `require-directories`     | #802|             |
| `require-engines`     | #868|             |
| `require-files`     | #803      |             |
| `require-funding`     | #804|             |
| `require-homepage`     | #865|             |
| `require-keywords`     | #866|             |
| `require-license`     | #846|     ✅        |
| `require-main`     | #805|             |
| `require-man`     | #806|             |
| `require-module`     | #807            |             |
| `require-name`     | #808|   ✅          |
| `require-optionalDependencies`     | #809|             |
| `require-os`     | #810|             |
| `require-peerDependencies`     | #811|             |
| `require-preferGlobal`     | #812|             |
| `require-private`     | #813|             |
| `require-publishConfig`     | #814|             |
| `require-repository`     | #867|             |
| `require-scripts`     | #815|             |
| `require-type`     | #1032  |      ✅       |
| `require-types`     | #816|             |
| `require-version`     | #817|    ✅         |

## Valid

| rule | issue | recommended |
|------|-------|-------------|
| `valid-author`     | #840|   ✅          |
| `valid-bin`     | #818|     ✅        |
|  `valid-bundleDependencies`    | #819|       ✅      |
| `valid-config`     | #820|       ✅      |
| `valid-cpu`     | #821|     ✅        |
| `valid-dependencies`     | #822|     ✅        |
| `valid-description`     | #823|      ✅       |
| `valid-devDependencies`     | #824|     ✅        |
| `valid-directories`     | #825 |       ✅      |
[...]

PR #1491, opened 2026-01-02
feat!: add require-repository rule

### LINKED ISSUE #867: 🚀 Feature: Bring in a require-repository rule
### Bug Report Checklist

### Overview

Following #42 -> #51: we've resolved to create a standalone `require-repository` rule. It should enforce that the field exists in the `package.json` file.

### Additional Info

💖

marcalexiei:
## PR Checklist

- [x] Addresses an existing open issue: fixes #867

## Overview

Add repository to properties listed in `src/rules/require-properties.ts`.

### DIFF
--- a/src/rules/require-properties.ts
+++ b/src/rules/require-properties.ts
@@ -20,7 +20,7 @@ const properties: [name: string, options?: CreateRequirePropertyRuleOptions][] =
 		["name", { ignorePrivateDefault: true, isRecommended: true }],
 		["optionalDependencies"],
 		["peerDependencies"],
-		["repository", { category: "Publishable" }],
+		["repository", { category: "Publishable", ignorePrivateDefault: true }],
 		["sideEffects", { category: "Publishable" }],
 		["type", { isRecommended: true }],
 		["types"],
--- a/docs/rules/require-repository.md
+++ b/docs/rules/require-repository.md
@@ -42,7 +42,7 @@ Examples of **correct** code for this rule:
 
 | Name            | Description                                                                                 | Type    | Default |
 | :-------------- | :------------------------------------------------------------------------------------------ | :------ | :------ |
-| `ignorePrivate` | Determines if this rule should be enforced when the package's `private` property is `true`. | Boolean | `false` |
+| `ignorePrivate` | Determines if this rule should be enforced when the package's `private` property is `true`. | Boolean | `true`  |
 
 <!-- end auto-generated rule options list -->
 


src/rules/require-properties.ts
```diff
@@ -20,6 +20,7 @@ const properties: [name: string, options?: CreateRequirePropertyRuleOptions][] =
 		["name", { ignorePrivateDefault: true, isRecommended: true }],
 		["optionalDependencies"],
 		["peerDependencies"],
+		["repository", { isRecommended: true }],
```
michaelfaith: We didn't have this earmarked for inclusion in the `recommended` config.  Just for reference, we listed all of the `require-` and `valid-` rules that we thought should go in the `recommended` config here: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/850

We still have more `require-` rules to add, so that we have the full set, but we don't have anymore that we think should go in the `recommended` config.  Just FYI for future contributions. 😊


```suggestion
		["repository"],
```

michaelfaith (changes requested):
Beautiful, thanks for another great contribution!  Just one request, and it should be good to go.

src/rules/require-properties.ts
marcalexiei: > Just for reference, we listed all of the require- and valid- rules that we thought should go in the recommended config here: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/850

Thanks, I'll check it before starting working on another `require-` rule issue.

---

It may be worth adding this to the publishable config, 
given the role of the repository field when publishing with provenance. 
This has already come up a few times:

- https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/867#issuecomment-2832672083
   
   > This would be useful, as it's a requirement to have the repository field when deploying with npm config provenance.
- https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/850#issuecomment-2833677703

   > I don't have a preference either way if this is included in the recommended config, but as 'require-repository' is a necessity for generating provenance statements, it could be a good standard to require people have this.

For now, I’ll remove it as requested, but I’m happy to add it back to the publishable configuration if needed.

michaelfaith: That's a good idea. We didn't have the publishable config at the time that that analysis was done. I'd  be in favor of adding it there 👍 (though I'm not sure the requirement for provenance still applies when using trusted publishing)

marcalexiei: > That's a good idea. We didn't have the publishable config at the time that that analysis was done. I'd be in favor of adding it there 👍

Done!

---

> I'm not sure the requirement for provenance still applies when using trusted publishing

I recall encountering an error related to the `repository` field while experimenting with provenance (haven't checked the behaviour with trusted publishing). 
That said, including a `repository` field still improves package discoverability 
by allowing users to locate the source repository directly, without needing to search for it externally.

src/rules/require-properties.ts
```diff
@@ -20,6 +20,7 @@ const properties: [name: string, options?: CreateRequirePropertyRuleOptions][] =
 		["name", { ignorePrivateDefault: true, isRecommended: true }],
 		["optionalDependencies"],
 		["peerDependencies"],
+		["repository", { category: "Publishable" }],
```
michaelfaith: Should we set ignorePrivateDefault to `true` for this, since we only care about it for published packages?

PR #1565, opened 2026-02-13
feat!: move the `publishable` require rules to `recommended`

### LINKED ISSUE #1562: 🚀 Feature: graduate simple require rules from `recommended-publishable` to `recommended`
### Feature Request Checklist

### Overview

Coming from https://github.com/flint-fyi/flint/pull/2141/changes#r2751740512 and #1560, there seems to be enough meaningful use cases for moving the `require-` rules that are currently unique to `recommended-publishable` into `recommended` and take advantage of the `ignorePrivate` option to only have them apply to publishable packages that way.  

### Additional Info

Once we do that and #1560 is implemented, there won't be anything left in the Publishable config that's unique... I'm inclined to leave it for now, even if it's the same as `recommended` (but adding a note in the docs explaining why), just in case there is a use case that comes up that would belong only in Publishable.  I feel like there are more areas of this space we haven't explored beyond the `require-` style rules, and I wouldn't want to deprecate it only to bring it back again once the use case arises.  So, I'm thinking of giving it at least a 6 month stay of execution after these are implemented and then re-asses.  @JoshuaKGoldberg what do you think about this piece specifically?  Would be interested in your take on this.

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1562

## Overview

This change moves all of the `require-` styles rules from `recommended-publishable` to the main `recommended` config, and uses their `ignorePrivate` property, instead, to apply them to "publishable" packages.  Following this change, there are no long any rules unique to `recommended-publishable`.

### DIFF
--- a/docs/rules/require-attribution.md
+++ b/docs/rules/require-attribution.md
@@ -172,7 +172,7 @@ Examples of **correct** code for this rule with default options:
 
 | Name                     | Description                                                               | Type    | Default |
 | :----------------------- | :------------------------------------------------------------------------ | :------ | :------ |
-| `ignorePrivate`          | Skip attribution requirements for packages with `"private": true`.        | Boolean | `false` |
+| `ignorePrivate`          | Skip attribution requirements for packages with `"private": true`.        | Boolean | `true`  |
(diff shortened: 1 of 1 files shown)


michaelfaith:
cc @ShayanTheNerd

PR #1682, opened 2026-03-18
fix(no-empty-fields): Allow `files` to be an empty list

### LINKED ISSUE #1679: 🐛 Bug: Conflicting rules for minimal packages
### Bug Report Checklist

### Expected

A package exporting a sole `index.js` file is valid.

### Actual

If a package only exports an `index.js`, the following rules conflict when it comes to declaring the `files` in package.json: `no-redundant-files`, `require-files`, `no-empty-fields`, and they are all present in the default `recommended` config

### Version

0.91.0

### ESLint Version

_No response_

### Additional Info

Hi :)

ClementValot:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1679

## Overview

Adding `files` as an exception to the `no-empty-fields` rule, since in some cases it can be correctly left empty (and not clash with `no-redundant-files`)

### DIFF
--- a/docs/rules/no-empty-fields.md
+++ b/docs/rules/no-empty-fields.md
@@ -42,10 +42,11 @@ Pass an array of top-level package properties to ignore.
 When provided, the rule won't report violations for the specified properties.
 This can be useful if you're using tools that take configuration from package.json and accept an empty array or object as valid non-default configuration.
 
-**Default:** `["files"]` 
+**Default:** `["files"]`
 
 > [!NOTE]
-> `files` is included by default since an empty array value is meaningfully different than not including it at all.  See [`no-redundant-files`](https://github.com/michaelfaith/eslint-plugin-package-json/blob/main/docs/rules/no-redundant-files.md) for more.
+> `files` is included by default since an empty array value is meaningfully different than not including it at all.
+> See [`no-redundant-files`](https://github.com/michaelfaith/eslint-plugin-package-json/blob/main/docs/rules/no-redundant-files.md) for more.
 
 Example of excluding the browserslist property used by [browserslist](https://www.npmjs.com/package/browserslist):
 


src/rules/no-empty-fields.ts
```diff
@@ -94,9 +94,14 @@ const getTopLevelProperty = (
...
 };
 
+// `files` can be empty since its contents can be inferred by `npm pack`
+const exceptions = ["files"];
+
 export const rule = createRule({
 	create(context) {
-		const ignoreProperties = context.options[0]?.ignoreProperties ?? [];
+		const ignoreProperties = new Set(
+			exceptions.concat(context.options[0]?.ignoreProperties ?? []),
```
michaelfaith: Maybe we should just default `ignoreProperties` to include `files` rather than having a separate exclusion mechanism?

ClementValot: I asked myself that!

This would force hardcoding `files` both in the default list line 103, and the `meta.defaultOptions.ignoreProperties` list line 136; causing a small duplication
Secondly, this would mix the user-defined exceptions from options, with the exceptions defined at the rule level. We default to never wanting to force `files` to be non-empty, because an empty list can always make sense there. 

(Only caveat is a user edge case where they would want to force it to be non-empty, but that would become a new option)

michaelfaith: > This would force hardcoding `files` both in the default list line 103, and the `meta.defaultOptions.ignoreProperties` list line 136; causing a small duplication

That's easy enough to get around by using a `const` to define it one place.  So, I'm not too fussed about the impact internally.  I'm more concerned with how the user's experience it.

> Secondly, this would mix the user-defined exceptions from options, with the exceptions defined at the rule level.

Can you explain this a bit more.  I'm not sure why those things can't be the same thing.  Can't the rule options both serve as a set of "reasonable" defaults for the rule, as well as a set that users can override with other values and it serve both needs?

> (Only caveat is a user edge case where they would want to force it to be non-empty, but that would become a new option)

This is my main concern.  If we hard-code it internally, there's no mechanism for users to override that.  I'm not sure what "that would become a new option" means here.  I wouldn't want to introduce another option on top of `ignoreProperties` to specifically bypass a built-in (hidden) default exclusion.  You know what I mean?

docs/rules/no-empty-fields.md
```diff
@@ -42,6 +42,8 @@ Pass an array of top-level package properties to ignore.
 When provided, the rule won't report violations for the specified properties.
 This can be useful if you're using tools that take configuration from package.json and accept an empty array or object as valid non-default configuration.
 
+Defaults to `["files"]`, since that field can be left empty as the files can be inferred; see `no-redundant-files`
```
michaelfaith:
```suggestion
**Default:** `["files"]` 

> [!NOTE]
> `files` is included by default since an empty array value is meaningfully different than not including it at all.  See [`no-redundant-files`](https://github.com/michaelfaith/eslint-plugin-package-json/blob/main/docs/rules/no-redundant-files.md) for more.
```

PR #1736, opened 2026-04-24
feat: add `restrict-top-level-properties` rule

### LINKED ISSUE #1435: 🚀 Feature: add a rule to control which extra top-level package.json properties are allowed
### Feature Request Checklist

### Overview

package.json allows arbitrary properties which are often used by other tools, such as `prettier`, `lint-staged` or `pnpm`. Some people prefer to consolidate different tools' settings in one place as much as possible to avoid creating a multitude of configuration files in the root of the repo - while others may prefer the opposite.

I have two use cases for which I would like a rule that can restrict certain top-level properties in `package.json`:
- The first case (described above): I prefer to keep my package.json as minimal and clean as possible, since tools' options sometimes can be long and/or unrepresentable in JSON;
- The second case: older versions of pnpm used the `pnpm` package.json key for pnpm specific options. Nowadays, the recommended way is to use a separate `pnpm-workspace.yaml` file, while the former method continues to work. It'd be nice to disallow `pnpm` property in that scenario.

### Proposed rule name

`restrict-top-level-properties`

### Proposed options

```ts
type Options = [{
  allow?: string[];
  ban?: (string | {property: string; message: string})[]; // Alternative option: Record<string, true | string>
}];
```

#### `allow`

- If omitted: all properties are allowed except those listed in `ban`.
- If an empty array: no extra top-level properties are allowed (only standard package.json fields).
- If a non-empty array: only the listed extra properties are allowed in addition to the standard fields.

`ban` can still be used alongside with `allow` to provide a custom message for specific properties.

If the same property appears in both `allow` and `ban`, the rule should throw an error.

#### `ban`

Provides a list of banned top-level properties with an optional custom message shown in the error message when the banned property is used. Can include any property, even standard ones like `name` or `version`. If the same property is banned multiple times, the last declaration takes precedence. (Alternatively, `ban` could be represented as `Record<string, true | string>` where a string value represents a custom message.)

### Additional Info

_No response_

andreww2012:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1435


docs/rules/restrict-top-level-properties.md
```diff
@@ -0,0 +1,171 @@
+# restrict-top-level-properties
+
+💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
+
+This rule allows you to disallow specific top-level properties in `package.json`.
+This is useful when you want to keep `package.json` minimal and enforce that certain tool configurations live in their own dedicated files.
+
+<!-- prettier-ignore-start -->
```
michaelfaith, line 10: Why are we disabling prettier here?

docs/rules/restrict-top-level-properties.md
```diff
@@ -0,0 +1,171 @@
...
+		],
+	},
+];
+```
+
+<!-- prettier-ignore-end -->
+
+## Options
+
+<!-- prettier-ignore-start -->
```
michaelfaith, line 35: I'm not sure why we're disabling prettier here too.  The doc generator is wired up with prettier, so they _shouldn't_ conflict with each other.  Were you seeing differently?

docs/rules/restrict-top-level-properties.md
```diff
@@ -0,0 +1,171 @@
+# restrict-top-level-properties
```
michaelfaith, line 1: Nice job on the documentation 👏

docs/rules/restrict-top-level-properties.md
```diff
@@ -0,0 +1,171 @@
...
+
+An array of entries describing which top-level properties are banned.
+Each entry is either:
+
+- A **string** — the property name to ban.
+- An **object** `{ property: string; message?: string }` — the property name to ban, with an optional custom message included in the error report.
+
+If the same property is listed multiple times, the last declaration takes precedence.
+
+Examples of **incorrect** code for this rule:
```
michaelfaith:
```suggestion
Examples of **incorrect** code for this rule, given the config above:
```

src/tests/rules/restrict-top-level-properties.test.ts
```diff
@@ -0,0 +1,329 @@
...
+	invalid: [
+		{
+			code: `{
+				"name": "test",
+				"prettier": { "semi": false }
+			}`,
+			errors: [
+				{
+					data: { customMessage: "", property: "prettier" },
+					messageId: "bannedProperty",
```
michaelfaith, line 15: Please include `line` numbers in all of the error objects.  After reworking the `valid-*` rules to have more granular violation reports, it's given me a greater appreciation for the precision of the reporting.  I've been wanting to add that as a requirement with the new v10 `RuleTester` options that are available, but then we'd also have to add `column` to like 800 tests, and I just haven't done that haha.

src/rules/restrict-top-level-properties.ts
```diff
@@ -0,0 +1,113 @@
...
+			category: "Best Practices",
+			description:
+				"Disallows specified top-level properties in package.json.",
+			recommended: false,
+		},
+		hasSuggestions: true,
+		messages: {
+			bannedProperty:
+				"The `{{ property }}` property is not allowed{{ customMessage }}",
+			removePropertySuggestion: "Remove the `{{ property }}` property.",
```
michaelfaith: Since we're reporting directly on the property, is there value in including the property name in the suggestion?  Perhaps just "Remove property" would be simpler and not lose any meaning.  What do you think?

src/rules/restrict-top-level-properties.ts
```diff
@@ -0,0 +1,113 @@
...
+			) {
+				for (const property of node.properties) {
+					if (!isJSONStringLiteral(property.key)) {
+						continue;
+					}
+
+					const propertyName = property.key.value;
+
+					if (!Object.hasOwn(banList, propertyName)) {
+						continue;
```
michaelfaith: I think the first `continue` is ok, since it's a quick bailout when the key isn't a string.  But since there aren't any other conditions that would cause us to skip this loop after this one, can we just inverse the condition and nest the remainder of the function within the `if`?  (similar reasoning as the conversation from the PJV PR)

src/rules/restrict-top-level-properties.ts
```diff
@@ -0,0 +1,113 @@
...
+		);
+
+		return {
+			"Program > JSONExpressionStatement > JSONObjectExpression"(
+				node: JsonAST.JSONObjectExpression,
+			) {
+				for (const property of node.properties) {
+					if (!isJSONStringLiteral(property.key)) {
+						continue;
+					}
```
michaelfaith, line 26: It looks like this condition isn't covered by unit tests.  Any way we can add coverage for this?

michaelfaith (changes requested):
Great start!  I love the way this is turning out. 🙌   Just a couple of minor things and request for additional unit test coverage.

src/rules/restrict-top-level-properties.ts
andreww2012: I do think there's still value in that. For example, when eslint is run from a command line, or any other issue reporter is used, it's more beneficial to see the actual property names than the same N error messages

michaelfaith: I agree for report / error messages. But where do you see suggestion text on the cli report?

andreww2012: I misread your comment a bit 🤦‍♂️Agreed that this is not necessary in a suggestion message

docs/rules/restrict-top-level-properties.md
andreww2012: For some reason, I'm getting different results when I run `pnpm format --write` and `pnpm docs:generate` 🤷‍♂️

michaelfaith (approved):
Sweet. Thanks again!  Nice work. 🥇

PR #1824, opened 2026-05-14
feat!: update node engines to `^22.22.2 || >=24.15.0`

michaelfaith:
This change bumps our node range to `^22.22.2 || >=24.15.0`.  Trying to get this just right before we cut 1.0.


JoshuaKGoldberg (approved):

andreww2012:
What was the reason for such tight requirements for 22 major version (`^22.22.2`)?

I see a lot of packages supporting only LTS versions of Node usually specify either of the following versions:
- `^22.11` - Node 22 marked as LTS;
- `^22.12` - added `require(esm)` support;
- `^22.18` - type stripping was enabled by default.

I can see the benefit of specifying `^22.22.x` for dev environment since 22.22 was a security release, but is there actually something in the package that would make it not properly work in the earlier Node versions?

michaelfaith:
> What was the reason for such tight requirements for 22 major version (`^22.22.2`)?

The new major version of `validate-npm-package-name`, which `package-json-validator` uses upstream increased their engines to that. So, i wanted to make sure we weren't going to be blocked on that update when the time comes.

PR #1850, opened 2026-05-18
docs: add missing (require|valid)-properties/ links to .md docs

JoshuaKGoldberg:
another quick lil docs fix I noticed we'd missed

### DIFF
--- a/docs/rules/require-attribution.md
+++ b/docs/rules/require-attribution.md
@@ -1,4 +1,4 @@
 # require-attribution
 
 > [!NOTE]
-> You can find the docs for this rule on our website: <https://eslint-plugin-package-json.dev/rules/require-attribution>
+> You can find the docs for this rule on our website: <https://eslint-plugin-package-json.dev/rules/require-properties/require-attribution>
--- a/docs/rules/require-attribution.md
+++ b/docs/rules/require-attribution.md
@@ -1,4 +1,4 @@
 # require-attribution
 
 > [!NOTE]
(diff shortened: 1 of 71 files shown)


docs/rules/require-attribution.md
```diff
@@ -1,4 +1,4 @@
 # require-attribution
 
 > [!NOTE]
-> You can find the docs for this rule on our website: <https://eslint-plugin-package-json.dev/rules/require-attribution>
+> You can find the docs for this rule on our website: <https://eslint-plugin-package-json.dev/rules/require-properties/require-attribution>
```
michaelfaith: `require-attribution` is the only rule with a `require-` prefix, that isn't part of the "simple" require- rules.  This one should be reverted 

screenshot

docs/rules/valid-repository-directory.md
```diff
@@ -1,4 +1,4 @@
 # valid-repository-directory
 
 > [!NOTE]
-> You can find the docs for this rule on our website: <https://eslint-plugin-package-json.dev/rules/valid-repository-directory>
+> You can find the docs for this rule on our website: <https://eslint-plugin-package-json.dev/rules/valid-properties/valid-repository-directory>
```
michaelfaith: This is also not part of the group of simple `valid-` rules. 

screenshot

michaelfaith (changes requested):
Ah, good catch. Thanks!  Just two that need to be reverted.

michaelfaith:
@JoshuaKGoldberg curious if you're still planning on landing this?

michaelfaith (approved):

PR #1857, opened 2026-05-18
chore: update eslint-doc-generator and remove patch

michaelfaith:
This change follows the merge of three changes I recently landed to `eslint-doc-generator`, which allows us to remove our local patch 🎉


michaelfaith:
This will need to wait a day for it to pass the supply chain security check in CI

PR #1880, opened 2026-05-27
ci: add e2e tests

### LINKED ISSUE #1879: 🛠 Tooling: create e2e tests for `/experimental` api
### Tooling Report Checklist

### Overview

We're currently smoke testing the main `plugin` just by proxy of having it in our repo's eslint config.  But now that we've added a new experimental version of the plugin, we have nothing that really tests it end to end.  And I especially concerned with the compat layer that sites between the entry points and the rules, configs, and plugins we're setting up for `@eslint/json` compatibility.  So, I'm proposing we set up a set of e2e tests for it.  

### Additional Info

These could either be temporary until we make the `experimental` api the primary api, or leave them there more long-term, since they could be testing a wider variety of scenarios.  We can make that call when the time comes.

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1879

## Overview

This change adds an e2e job to our CI workflow.  The new `e2e` directory now contains a new pnpm workspace, with a sub-directory for each e2e test.  That test dir should have a `package.json` and `eslint.config`.  The `package.json` should have a `lint` command.  The `e2e/package.json` will execute the lint command on all workspaces concurrently.  The two e2e tests I've set up for now, are testing the `configs` and `rules` from the new `experimental` entry point.

Also, I have `e2e` depending on the `build` step, so the build doesn't have to run twice.

screenshot

screenshot

### DIFF
--- a/e2e/experimental-configs/package.json
+++ b/e2e/experimental-configs/package.json
@@ -1,0 +1,28 @@
+{
+  "name": "experimental-configs",
+  "version": "1.1.0",
+  "description": "e2e tests for eslint-plugin-package-json's experimental configurations.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/michaelfaith/eslint-plugin-package-json.git",
+    "directory": "e2e/experimental-configs"
+  },
+  "license": "MIT",
+  "author": {
+    "name": "michael faith",
+    "url": "https://michael.faith"
+  },
+  "sideEffects": false,
+  "type": "module",
+  "exports": {
+    "./package.json": "./package.json"
+  },
+  "files": [],
+  "scripts": {
+    "lint": "eslint ."
+  },
+  "devDependencies": {
+    "eslint": "10.4.0",
+    "eslint-plugin-package-json": "file:../.."
+  }
+}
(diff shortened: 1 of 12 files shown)


e2e/experimental-configs/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "1.1.0",
+  "description": "e2e tests for eslint-plugin-package-json's experimental configurations.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/michaelfaith/eslint-plugin-package-json.git",
+    "directory": "e2e/experimental-configs"
+  },
+  "license": "MIT",
+  "author": {
+    "name": "michael faith",
```
JoshuaKGoldberg, line 12: In case this matters to you 😜 

```suggestion
    "name": "Michael Faith",
```

JoshuaKGoldberg (approved):
Yay, thoroughness!

e2e/experimental-configs/package.json
michaelfaith: It does, and that's why it's lower case haha 😆

PR #1892, opened 2026-05-29
feat: add `prefer-scoped-name` rule

### LINKED ISSUE #1875: 🚀 Feature: rule `prefer-scoped-name`
### Feature Request Checklist

### Overview

I will be implementing a "safe list" rule for scopes which we support folks publishing with and wanted to propose I do so here instead of on top of this plugin in our internal code base. Something like this:

```
rules: {
  // Require a single scope, any unrecognized will be fixed to this value
  'package-json/prefer-scoped-name": ['error', { default: '@foo' }],

  // Prefer the @foo scope, any unrecognized will be fixed to this value
  // also allow the @bar scope which will neither error or be fixed
  'package-json/prefer-scoped-name": ['error', {
    default: '@foo',
    allowed: ['@bar']
  }]
}
```

wesleytodd:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1875

## Overview

Adds `prefer-scope` which is a "safe list" rule for scopes.

```
rules: {
  // Require a single scope, any unrecognized will be fixed to this value
  'package-json/prefer-scoped-name": ['error', { default: '@foo' }],

  // Prefer the @foo scope, any unrecognized will be fixed to this value
  // also allow the @bar scope which will neither error or be fixed
  'package-json/prefer-scoped-name": ['error', {
    preferred: '@foo',
    allowed: ['@bar']
  }]
}
```


src/tests/rules/prefer-scoped-name.test.ts
michaelfaith, line 1: We'll definitely need a lot more tests than this.  We need to test every combination of options in both the `valid` and `invalid` group.  We also want to test cases where `name` isn't present at all (like `{}`), to make sure it's only applying to the `name` property.  Also cases where `name` is a nested property under some other top-level property, to ensure it's only applying to the top-level `name`.  Etc.

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,102 @@
...
+    return {
+      'Program > JSONExpressionStatement > JSONObjectExpression': (node: JsonAST.JSONObjectExpression) => {
+        const exp = node.properties.find((prop) => {
+          return prop.key.type === 'JSONLiteral' && prop.key.value === 'name';
+        });
+        if (!exp) {
+          return;
+        }
+
+        // @ts-expect-error exp.value.value is a string despite the types being wrong for JSONArrayExpression
```
michaelfaith: There's nothing above to narrow the type.  I'm not sure why the comment mentions array expressions here, since this _should_ be a `JSONStringLiteral`, but there's been no attempt to validate that the value is infact a string literal.  What if someone happens to have an `array` for the `name` property to for some reason.  There's nothing preventing this code from blowing up.

```json
{
  "name": []
}
```
should not be touched by this rule.  We should have a test for that as well.

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    if (!preferred) {
+      return {};
+    }
+
+    return {
+      'Program > JSONExpressionStatement > JSONObjectExpression': (node: JsonAST.JSONObjectExpression) => {
+        const exp = node.properties.find((prop) => {
+          return prop.key.type === 'JSONLiteral' && prop.key.value === 'name';
+        });
```
michaelfaith: For all the other rules that operate on only one of the top level properties, we've generally built that into the query selector string.  Please use that pattern here as well.

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,102 @@
...
+          });
+        }
+      }
+    };
+  },
+  meta: {
+    defaultOptions: [[]],
+    docs: {
+      category: 'Best Practices',
+      description: 'Prefer a scoped package name'
```
michaelfaith:
```suggestion
      description: 'Enforce that `name` uses a specific scope or set of scopes.'
```

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,102 @@
...
+    },
+    fixable: 'code',
+    messages: {
+      missingPreferredScope: 'Use the preferred scope {{preferred}} in package name, found {{name}}',
+      useAllowedScope: 'Use an allowed scope (preferably {{preferred}}), found {{scope}} in {{name}}',
+      usePreferredScope: 'Use the preferred scope {{preferred}}, found {{scope}} in {{name}}'
```
michaelfaith:
```suggestion
      missingPreferredScope: 'Use the preferred scope "{{preferred}}" in package name; found "{{name}}"',
      useAllowedScope: 'Use an allowed scope (preferably "{{preferred}}"); found "{{scope}}"',
      usePreferredScope: 'Use the preferred scope "{{preferred}}"; found "{{scope}}"'
```

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,102 @@
...
+              }
+            ],
+            description: 'Allowed scopes'
+          },
+          preferred: {
+            description: 'Prefered scope',
+            type: 'string'
+          }
+        },
+        requiredProperties: ['preferred'],
```
michaelfaith: In https://github.com/michaelfaith/eslint-plugin-package-json/issues/1875#issuecomment-4549200300, I had `preferred` (then `default`) also marked as optional.  We landed on having at least _some_ option required, but it could be either one.  I.e. a user could just provide `allowed` and it would just check that the scope they're using matches one of the allowed scopes (and there'd be no auto-fix).

michaelfaith (changes requested):
Thanks for the contribution!  Nice start, but a number of issues that'll need to be worked out.  Also, several of the CI steps are failing (did you happen to run the unit tests locally?).  If you haven't already, our [DEVELOPMENT.md](https://github.com/michaelfaith/eslint-plugin-package-json/blob/main/.github/DEVELOPMENT.md), is usually a good doc to review before making a first contribution.  It includes local development commands.

michaelfaith:
Additional more general feedback: this doesn't add the rule to the plugin, and there isn't a doc page for it, which will need to be added before we can move forward with the change.

src/rules/prefer-scoped-name.ts
wesleytodd: Like this you mean? Just making sure I understand the comment.

```suggestion
'Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.value=name]'
```

src/rules/prefer-scoped-name.ts
wesleytodd: The type error included this being a missing property on `JSONArrayExpression`. 

```
Property 'value' does not exist on type 'JSONArrayExpression'
```

src/rules/prefer-scoped-name.ts
wesleytodd: Sounds good. Will make that change when I get back to this.

src/tests/rules/prefer-scoped-name.test.ts
wesleytodd: Yep, my intent was to just get started. I usually only have an hour or so each morning to do some oss work, I probably should have mentioned I didn't see this PR as complete but more of a start for the conversations to make sure I wasn't going in a bad direction you wouldn't want.

michaelfaith: I see.  In the future, I'd recommend just leaving it as Draft then, if it's not ready for a full / thorough review.  I'm happy to do a less detailed review of a draft pr, if you ask for it, but won't call out things like this.

src/rules/prefer-scoped-name.ts
michaelfaith: Because you haven't narrowed the type yet.  The type is too broad right now.  There needs to be some "kind" checking above this to ensure the thing you're working with is of the node type you expect.  We do a lot of it in other rules, and I'd encourage you to use existing rules as a point of reference. With that said, if you do what I suggested in https://github.com/michaelfaith/eslint-plugin-package-json/pull/1892/changes/BASE..a397915cd4e1ed8874ac9b43f465a675f2311e5c#r3328758890 (which also has plenty of examples around the repo), the need to narrow the type is reduced.

src/rules/prefer-scoped-name.ts
michaelfaith: Yep 👍, and you'll also need to adjust the type of `node`.  There are examples of that in other rules as well.

src/tests/rules/prefer-scoped-name.test.ts
wesleytodd: Will do! Honestly just habit to click the button without much thought. Converted to draft.

src/rules/prefer-scoped-name.ts
wesleytodd: Made the proper schema change, but the lint rules don't like this `anyOf` because they don't have descriptions. I just left this dummy value, if you have a suggestion on how you want this resolved please let me know.

src/tests/rules/prefer-scoped-name.test.ts
wesleytodd: Added a few more tests, but my time was overshot today. I think there are a few more things to test and just wanted to push what I had. If you have specific things you want me to add tests for please call them out. I will add more of what I think is important to test next time I get to this otherwise.

src/rules/prefer-scoped-name.ts
wesleytodd: I used most of my time fighting the tooling setup in this project instead of actually fixing anything. I probably am going to have to abandon this if it is not close to a merge-able state after what I just pushed. So while I left one ts-ignore for the complaint about `npm-package-arg`'s type I think I did what you are suggesting with the narrowing via the filter. Let me know if I missed the mark here, but without more clear guidance I did what I had time for.

wesleytodd:
> this doesn't add the rule to the plugin

fixed.

> and there isn't a doc page for it, which will need to be added before we can move forward with the change.

Please point me in the right direction for adding this. As I said in my other comments I have used all the available time I have for this contribution addressing the inconsistent tooling complaints. If I am going to spend any more time on this I need clear instructions on how to get this across the finish line.

src/tests/rules/prefer-scoped-name.test.ts
michaelfaith: I believe https://github.com/michaelfaith/eslint-plugin-package-json/pull/1892#discussion_r3328747169 covers most of what I was thinking.

src/rules/prefer-scoped-name.ts
michaelfaith: > I used most of my time fighting the tooling setup in this project

Out of curiosity what tooling issues did you have?

michaelfaith:
> If I am going to spend any more time on this I need clear instructions on how to get this across the finish line.

It's all in our [development guidelines](https://github.com/michaelfaith/eslint-plugin-package-json/blob/main/.github/DEVELOPMENT.md#generating-docs).  I get the sense that you didn't read those, since most of the things you've said you need direction on are already covered there. 😅  That command will generate a skeleton doc page, that you then would fill out the rest.

src/tests/rules/prefer-scoped-name.test.ts
wesleytodd: Ah cool, I missed two:

1. name isn't present at all
2. name is a nested property under some other top-level property

Will add those.

src/rules/prefer-scoped-name.ts
wesleytodd: The linting rules required `defaultOptions` (and that random description in a place that doesn't make sense) but since this doesn't have any passing `[]` caused the typechecking on the schema to fail. The intent of the code is correct, the linting is probably what is wrong not the typechecking since this rule is not intended to ever be on by default it is intentional to have no default options but *require* options when it is in use.

That was the main one. There was also some silent failure in the precommit hooks, but I just `--no-verify` for that. When I ran it it assumed a global `pnpm` install which I don't do, but that didn't seem to be the silent failure. I didn't debug that further.

And lastly, I just dislike when typescript complains about correct code and makes you write code that ends with worse javascript at runtime. This is just my complaining because I am an old man shaking my fist at the clouds, nothing setup incorrectly if TS is something you like. But I have a choice on how I spend my little time available for OSS and I don't contribute to TS projects for a reason. Likely that means I will not become a regular contributor to this library despite it being most of what I need. I am probably going to write my rules outside of the plugin and use a config package to bundle them together from now on.

EDIT: I can't tell (and didn't go digging) if it is the abstraction from this repo or not, but also the fact that the tests don't clearly tell me what test case failed in the stack trace but that I have to infer it from the message wasn't helping.

wesleytodd:
I did read them when I originally cloned the repo, but thanks for the link, that is what I was asking for as a pointer on where to start with that.

wesleytodd:
I want to follow up since as I was thinking about this interaction today I realized it must seem like I was either being an entitled lazy user or an adversarial contributor. Yes I wish I could have just dropped in the code I had written at work as is, but my feedback here was more to share my experience as a contributor who didn't have great success with my first contribution. I hope that my feedback can be constructive.

If I am the only one who has had trouble with their first contribution in this way, awesome! Please disregard my feedback. Either way, I don't expect any change or really anything from you. It is your project, and it is a good one (or I wouldn't be here hoping to contribute and starting my journey as a user). I just don't want to waste the opportunity to share the struggles (even minor) that I had as a first time contributor.

It is much more an effect of me trying to carve out some time to contribute and not having much time to do so than anything about this project or the setup. I just want to maximize the couple of hours a week I can put on this and being frustrated when that is not enough to put together a complete/reasonable PR for reasons not directly related to the feature I was contributing is not fun OSS. I hope this helps clarify what happened here and avoids any interpersonal tensions from my style of direct feedback.

wesleytodd:
I think I addressed the feedback. This is probably ready for a review now?

src/rules/prefer-scoped-name.ts
michaelfaith: > The intent of the code is correct, the linting is probably what is wrong not the typechecking since this rule is not intended to ever be on by default it is intentional to have no default options but require options when it is in use.

That's a bit of a misunderstanding of how `defaultOptions` work in ESLint.  `defaultOptions` doesn't in any way relate to whether or not a rule is enabled.  It just let's ESLint know that _when_ a rule is enabled, and the user hasn't provided any options, what options should be used by default. 

>  the linting is probably what is wrong

This rule isn't unique to our repo.  This is a rule provided in the `recommended` config of `eslint-plugin-eslint-plugin`, which is a plugin developed by the eslint community specifically focused on enforcing common conventions and standards for plugin authors.  This is the specific rule you're talking about: https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-default-options.md.  That plugin is widely used, and I would be willing to wager that all of your favorite ESLint plugins probably have some set of its rules enabled in their repo.  We're just using what's in their `recommended` config. 

This is more general advice, than anything specific to this change: rather than rushing to judgement and getting frustrated about something not working quite the way you'd expect, be open to the idea that maybe you don't have all of the information (which is totally ok, and nobody would expect you to, if you don't work in this space a lot).  Otherwise you're closing the door to a possible learning opportunity. I recognize and appreciate that you're doing this all in your spare time, but so are we. 😊

site/src/content/docs/rules/prefer-scoped-name.md
```diff
@@ -0,0 +1,88 @@
...
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+This rule enforces that the package `name` uses a scope (ex. `@scope/name`).
+
+## Rule Details
+
+This rule is not included in any default configurations because it requires custom configuration based on your project.
```
michaelfaith, line 14: This line can be removed.  The header would include any configs it's in, if it were in a config.

site/src/content/docs/rules/prefer-scoped-name.md
```diff
@@ -0,0 +1,88 @@
...
+Configure either a single preferred scope, multiple allowed scopes, or combine them if you have one preferred scope but
+allow for others without causing a lint error.
```
michaelfaith, line 16:
```suggestion
You can configure either a single preferred scope, multiple allowed scopes, or combine them if you have one preferred scope but allow for others.
```

site/src/content/docs/rules/prefer-scoped-name.md
```diff
@@ -0,0 +1,88 @@
...
+    "error",
+    {
+      "preferred": "@foo",
+      "allowed": [ "@bar", "@baz" ]
+    }
+  ]
+}
+```
+
+<!-- end auto-generated rule options list -->
```
michaelfaith, line 88: This line should be directly below the table (with a space between), so the doc generator will only adjust the table, rather than all of the content you added yourself.

site/src/content/docs/rules/prefer-scoped-name.md
```diff
@@ -0,0 +1,88 @@
...
+| Name        | Description    | Type   |
+| :---------- | :------------- | :----- |
+| `allowed`   | Allowed scopes | String or String[] |
+| `preferred` | Prefered scope | String |
+
+
+### `allowed`
+
+Allows a single scope or list of scopes in named packages. This option will never be used in a fix, and will only error
+if a scope is used that is not in the allowed list. With only this option, un-scoped names are allowed.
```
michaelfaith, line 31:
```suggestion
Allows a single scope or list of scopes in named packages. 
This option will never be used in a fix, and will only error if a scope is used that is not in the allowed list. With only this option, un-scoped names are allowed.
```

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+        anyOf: [
+          {
+            description: 'this lint rule doesn\'t make sense',
+            required: ['preferred']
+          },
+          {
+            description: 'this lint rule doesn\'t make sense',
+            required: ['allowed']
+          }
+        ],
```
michaelfaith, line 117: So, this is an odd way to do this, and probably why the lint rule is throwing a fit.  Normally, you'd have a top level `oneOf` that acts as a sort of type union that has two different variations of the schema, one with `allowed` required, and the other with `preferred` required.  I wonder if it would just be easier to make both options optional, use `[{}]` for the `defaultOptions`, and then check that one is enabled within the rule itself (which you're already doing).  Not a prescription, but just an idea to make it simpler.  `restrict-dependency-ranges` is an example where we've dynamically constructed a complex schema like I'm talking about using bits that are the same in multiple places.

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+        }
+      }
+    };
+  },
+  meta: {
+    // This rule does not have defaults but also requires one of those
+    // properties. I think this is also a conflict of the eslint rules
+    // with they typescript checks
+    // eslint-disable-next-line eslint-plugin/require-meta-default-options
+    defaultOptions: [],
```
michaelfaith, line 96:
```suggestion
    defaultOptions: [{}],
```
(If you go with my suggestion below)

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+export const rule = createRule({
+  create(context) {
+    const { allowed: _allowed, preferred } = (context.options[0] ?? {}) as { allowed: string | string[], preferred: string};
+
+    // convert single allowed string scope to array
+    let allowed: null | string[] = null;
+    if (_allowed && !Array.isArray(_allowed)) {
+      allowed = [_allowed];
+    } else if (_allowed) {
+      allowed = _allowed as string[];
```
michaelfaith, line 16: I'm surprised that the conditions you've setup don't narrow the type to `string[]` already.  Can you confirm that this `as` casting is still necessary?

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
+import type { AST as JsonAST } from 'jsonc-eslint-parser';
+
+import npa from 'npm-package-arg';
+
+import { createRule } from "../createRule.ts";
+
+export const rule = createRule({
+  create(context) {
+    const { allowed: _allowed, preferred } = (context.options[0] ?? {}) as { allowed: string | string[], preferred: string};
```
michaelfaith, line 9: You shouldn't really need to `as` cast here, if the `schema` is set up properly.  `createRule` will set the types up for you.  If it's necessary, then it's likely the schema isn't set up right.

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (!preferred && !allowed?.length) {
+      return {};
+    }
+
+    return {
+      'Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.type=JSONLiteral][value.type=JSONLiteral][key.value=name]': (node: JsonAST.JSONProperty) => {
+        let pkgarg;
+        try {
+          // @ts-expect-error needs some type narrowing
+          pkgarg = npa(node.value.value as string);
```
michaelfaith, line 28: You can use something like this, instead, which will avoid needing to cast and `ts-expect-error` here

https://github.com/michaelfaith/eslint-plugin-package-json/blob/b0b83951c4b84da5436b0b2c0636f4ea1cd656cf/src/rules/restrict-dependency-ranges.ts#L126-L129

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+        } catch {
+          // what to do with an invalid package name? I don't see any other
+          // rule dealing with this case.
```
michaelfaith, line 31: `valid-name` covers this.

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        // We have a preferred scope but no scope was specified
+        if (preferred && !pkgarg.scope) {
+          context.report({
+            data: {
+              name: pkgarg.name,
+              preferred
+            },
+            fix: (fixer) => {
+              // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
michaelfaith, line 43: Why are we disabling here?

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        // We have a preferred scope that doesn't match and no other allowed scopes to check
+        if (pkgarg.scope !== preferred && !allowed?.length) {
+          context.report({
+            data: {
+              preferred,
+              scope: pkgarg.scope
+            },
+            fix: (fixer) => {
+              // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
michaelfaith, line 60: Shouldn't need to disable here.

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+              return fixer.replaceText(node.value, `"${(pkgarg.name || '').replace((pkgarg.scope || ''), preferred)}"`);
+            },
+            messageId: 'usePreferredScope',
+            node: node.value,
+          });
+          return;
+        }
+
+        // Neither preferred scope or allowed scopes match
+        // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
michaelfaith, line 70: Here too

src/rules/prefer-scoped-name.ts
```diff
@@ -0,0 +1,142 @@
...
+          context.report({
+            data: {
+              preferred: preferred || allowed.join(', '),
+              scope: pkgarg.scope
+            },
+            fix: (fixer) => {
+              if (!preferred) {
+                return null;
+              }
+              // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
michaelfaith, line 81: Here too

michaelfaith (changes requested):
Great progress. Thanks for continuing the work on it.  I've left a handful of feedback.  I also noticed we're missing four or five branches and one line entirely in the codecov report.  Recommend at least adding enough tests to cover the branches.  The line in the catch block, I don't feel particularly strong about.

PR #1936, opened 2026-06-15
feat(sort-collections): add custom sort order option

### LINKED ISSUE #1390: 🚀 Feature: `sort-collections` should sort all keys
### Feature Request Checklist

### Overview

As per the changes to #1374, I would now like for *all* keys, regardless of where they come from, to be sorted.

I believe we should do three things:

1. Set the following code to only run when specific dependencies are sorted
```ts
if (!toSort.has(key)) {
	return;
}
```

2. Sort all objects (regardless of nesting) lexicographically

3. This may be more of a nice to have (and maybe out of the scope of this request), I'd like to be able to supply specific dependency sort order. Like how `order-properties` works, I'd like to be able to supply a specific key (e.g. "nx"), and then provide the reference for what order this is sorted in.

We could provide nested keys to this, but any key not provided would default back to lexicographical order. For example though, with nx, I'd like to supply a custom order that matches the [nx schema](https://raw.githubusercontent.com/nrwl/nx/refs/heads/master/packages/nx/schemas/nx-schema.json) order, rather than being in alphabetical.

### Additional Info

Option #3 may be better for another feature request. I believe it may also be a breaking change to implement (option wise) as the current options *only* support an array of string, and although we could change it to an array of strings and objects, the semantics of it would be confusing. E.g. if I had:

```json
{
	"package-json/sort-collections": [
		"error",
		[{ nx: { ... my custom order }]
	]
}
```

Am I now saying I only want nx formatted? This isn't the case.

christopher-buss:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1390

## Overview

Allow an option entry to be an object `{ key, order }` so a collection can be sorted by a specified reference order (e.g. nx schema) instead of lexicographically. Keys present in the collection but not listed in order are appended after the listed keys, sorted lexicographically among themselves. Collections not named in the options are left untouched (unchanged).

### DIFF
--- a/site/src/content/docs/rules/sort-collections.mdx
+++ b/site/src/content/docs/rules/sort-collections.mdx
@@ -141,9 +141,23 @@ Incorrect vs. correct ordering of a complete lifecycle group:
 
 ## Options
 
+{/* begin auto-generated rule options list */}
+
+| Name    | Description                                                                                 | Type     | Required |
+| :------ | :------------------------------------------------------------------------------------------ | :------- | :------- |
+| `key`   | The collection property to sort.                                                            | String   | Yes      |
+| `order` | The order to sort the collection by. Keys not listed are appended in lexicographical order. | String[] | Yes      |
+
+{/* end auto-generated rule options list */}
+
 Pass an array of package properties to require sorting on those collections.
 To specify a nested property, use dot notation.
 All of their values must be objects.
+
+Each entry is either:
+
+- A string: the collection to sort the default way (lexicographically, or lifecycle-aware for `scripts`)
+- An object (`{ key: string; order: string[] }`): sort the collection named by `key` using the given `order`; keys not listed in `order` are appended afterwards, sorted lexicographically
 
 Example of only sorting `devDependencies` and `pnpm.patchedDependencies`:
 
@@ -152,6 +166,17 @@ Example of only sorting `devDependencies` and `pnpm.patchedDependencies`:
   "package-json/sort-collections": [
     "error",
     ["devDependencies", "pnpm.patchedDependencies"]
+  ]
+}
+```
+
+Example of enforcing a custom order for `nx` while still sorting `dependencies` the default way:
+
+```json
+{
+  "package-json/sort-collections": [
+    "error",
+    ["dependencies", { "key": "nx", "order": ["npmScope", "affected"] }]
   ]
 }
 ```
--- a/src/rules/sort-collections.ts
+++ b/src/rules/sort-collections.ts
@@ -59,50 +59,45 @@ export const rule = createRule({
         }
 
         const currentOrder = collection.properties;
-        let desiredOrder: JsonAST.JSONProperty[];
 
         const isScripts = keyPartsReversed.at(-1) === 'scripts';
 
-        if (isScripts && !customOrder) {
-          // For scripts we'll use `sort-package-json`
+        // The "natural" order for keys not pinned by a custom order: `scripts`
+        // are lifecycle-aware (via `sort-package-json`), everything else is
+        // lexicographical.
+        let naturalCompare: (a: string, b: string) => number;
+        if (isScripts) {
           const scriptsSource = context.sourceCode.getText(node);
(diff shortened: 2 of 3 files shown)


christopher-buss:
apparently the bot doesn't like [X] over [x]

christopher-buss:
The coverage not being 100% here is because of the eslint < 9.15 not applying defaultOptions - and seems to be pre-existing here. I can add a test to cover this if needed 🙂

site/src/content/docs/rules/sort-collections.mdx
```diff
@@ -141,10 +141,24 @@ Incorrect vs. correct ordering of a complete lifecycle group:
...
 
+{/* begin auto-generated rule options list */}
+
+| Name    | Description                                                                                 | Type     | Required |
+| :------ | :------------------------------------------------------------------------------------------ | :------- | :------- |
+| `key`   | The collection property to sort.                                                            | String   | Yes      |
+| `order` | The order to sort the collection by. Keys not listed are appended in lexicographical order. | String[] | Yes      |
+
+{/* end auto-generated rule options list */}
+
```
michaelfaith, line 152: Thanks for adding this 👍

src/tests/rules/sort-collections.test.ts
```diff
@@ -220,6 +220,54 @@ ruleTester.run('sort-collections', rule, {
...
+        },
+      ],
+      filename: 'package.json',
+      options: [[{ key: 'nx', order: ['npmScope', 'affected'] }]],
+      output: `{
+	"nx": {
+    "npmScope": "test",
+    "affected": {}
+  }
+}`,
```
michaelfaith, line 247: [No action needed] Just filed #1937, to fix the indention issue we have in this rule (which pre-dates this PR).

src/tests/rules/sort-collections.test.ts
```diff
@@ -220,6 +220,54 @@ ruleTester.run('sort-collections', rule, {
   }
 		}
 	}
+}`,
+    },
+    // custom order: listed keys follow the specified order
+    {
```
michaelfaith, line 226: nit: would you mind adding a `name` to the new test cases?  Anytime test cases have options or settings defined, I like to have `name`'s defined also, since the default test output only shows the `code` if no name is defined.  And for test cases that only differ by their options / settings, it can be hard to pinpoint which test case failed.  It can be what you have in the comment above...

src/tests/rules/sort-collections.test.ts
michaelfaith, line 1: Test cases are solid.  I'd like to see a case of overriding the order of `scripts` though, if you wouldn't mind adding a test case for that too.

src/rules/sort-collections.ts
```diff
@@ -49,14 +51,19 @@ export const rule = createRule({
...
           return;
         }
 
         const currentOrder = collection.properties;
         let desiredOrder: JsonAST.JSONProperty[];
 
-        if (keyPartsReversed.at(-1) === 'scripts') {
+        const isScripts = keyPartsReversed.at(-1) === 'scripts';
+
+        if (isScripts && !customOrder) {
```
michaelfaith: What if someone provides a custom order for `scripts`, but has keys that aren't included in the custom order config?  I would expect the remaining keys to use a similar lifecycle aware behavior as they have when no custom order is provided.  But as the logic stands now, it'll fall into the lexicographic sorting for those remaining keys.

michaelfaith (changes requested):
Great start!  Thanks for the contribution.  Overall, I'm pretty happy with the direction you're going here.  Other than some minor nits and an ask for more test cases, the main piece of feedback i have is around the behavior of `scripts` sorting when a custom order is provided for `scripts`.  Right now, any remaining keys that aren't specified in the custom order config would be sorted lexicographically, rather than being life-cycle aware, which i don't think is correct.  More generally: any unspecific keys for a property that has a custom order applied, should be sorted the way they normally would after the custom order.  In the case of `scripts` that's life-cycle aware.

michaelfaith (approved):
This is good stuff; nice work! 🚀  Appreciate the contribution

PR #2001, opened 2026-07-10
feat: deprecate `recommended-publishable` config

### LINKED ISSUE #1843: 💬 RFC: deprecate `recommended-publishable` config
### RFC Checklist

### Overview

Right now, the `recommended-publishable` config is identical to `recommended`.  Originally, the intent was to cater that config to the use case of publishable packages.  But with how we've integrated the `ignorePrivate` option, it allowed us to make all of the rules that were unique to `recommended-publishable` just part of the regular `recommended` config.  I was hesitant to kill it entirely for the v1 release, since there may be rules we haven't thought of yet, that would be a good fit for it.  And I'm still open to that, if anyone has any ideas.  If we haven't thought of any by the time this RFC closes, I'd propose we mark the config as `deprecated`, remove it from our docs, and queue it up for removal in our next major version.

Thoughts?

### This RFC will remain open until

2026-07-13

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1843

## Overview

This change deprecates the `recommended-publishable` config, now that we've consolidated around the `recommended` config for everything.  After baking in awareness for when a package is private or not, the `recommended` config became behaviorally the same as `recommended-publishable`, making it redundant.

Please move to the `recommended` config.


src/plugin.ts
```diff
@@ -98,6 +98,7 @@ export const plugin = {
       },
       rules: recommendedRules,
     },
+    /** @deprecated use the recommended config instead. this config will be removed in v2 */
```
michaelfaith:
```suggestion
    /** @deprecated use the recommended config instead. this config will be removed in a future major release */
```

PR #2051, opened 2026-07-29
fix(no-local-dependencies): exempt `bundle(d)Dependencies` from check

### LINKED ISSUE #2048: 🐛 Bug: `no-local-dependencies`: should ignore dependencies listed in `bundle(d)Dependencies`
### Bug Report Checklist

### Expected

Prod dependencies using the `file:` protocol are correctly resolved by almost all package managers (except for the Yarn classic) during installation, if they are present in `bundle(d)Dependencies` array.

That's how [I've been declaring some dependencies in my package](https://npmx.dev/package-code/eslint-config-un/v/1.0.0-beta.12/package.json#L94) for some time (they were patched).

So such entries should not be flagged by this rule.

### Actual

Currently `foo` dependency is reported:

```json
{
  "dependencies": {
    "foo": "file:foo.tgz"
  },
  "bundleDependencies": ["foo"]
}
```

### Version

1.6.2

### ESLint Version

10.8.0

### Additional Info

_No response_

andreww2012:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2048

### DIFF
--- a/site/src/content/docs/rules/no-local-dependencies.mdx
+++ b/site/src/content/docs/rules/no-local-dependencies.mdx
@@ -10,6 +10,11 @@ import { TabItem, Tabs } from '@astrojs/starlight/components';
 This rule enforces that no `dependencies` are declared with local file paths, since this will likely result in errors when installing from a registry.
 This includes any paths that start with `file:`, `link:`, `./`, or `../`.
 
+Dependencies listed in [`bundleDependencies`](https://docs.npmjs.com/cli/v12/configuring-npm/package-json#bundledependencies) (or its `bundledDependencies` alias) are exempt, since their contents are published within the package's own tarball and so remain resolvable when installed from a registry.
+Setting `bundleDependencies` to `true` bundles every dependency, and likewise exempts them all.
+
+If both spellings are present, only `bundleDependencies` is honored, matching how npm resolves the alias.
+
 ## Examples
 
 <Tabs>
@@ -35,6 +40,15 @@ This includes any paths that start with `file:`, `link:`, `./`, or `../`.
     "bar": "^1.2.3",
     "baz": "~0.1.2"
   }
+}
+```
+
+```json
+{
+  "dependencies": {
+    "foo": "file:foo.tgz"
+  },
+  "bundleDependencies": ["foo"]
 }
 ```
 
--- a/site/src/content/docs/rules/no-local-dependencies.mdx
+++ b/site/src/content/docs/rules/no-local-dependencies.mdx
@@ -13,7 +13,7 @@ This includes any paths that start with `file:`, `link:`, `./`, or `../`.
 Dependencies listed in [`bundleDependencies`](https://docs.npmjs.com/cli/v12/configuring-npm/package-json#bundledependencies) (or its `bundledDependencies` alias) are exempt, since their contents are published within the package's own tarball and so remain resolvable when installed from a registry.
 Setting `bundleDependencies` to `true` bundles every dependency, and likewise exempts them all.
 
-If both spellings are present, only `bundleDependencies` is honored, matching how npm resolves the alias.
+The `bundledDependencies` alias is only consulted when `bundleDependencies` is absent or falsy, matching how npm resolves the two.
 
 ## Examples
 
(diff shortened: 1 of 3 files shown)


site/src/content/docs/rules/no-local-dependencies.mdx
```diff
@@ -10,6 +10,11 @@ import { TabItem, Tabs } from '@astrojs/starlight/components';
 This rule enforces that no `dependencies` are declared with local file paths, since this will likely result in errors when installing from a registry.
 This includes any paths that start with `file:`, `link:`, `./`, or `../`.
 
+Dependencies listed in [`bundleDependencies`](https://docs.npmjs.com/cli/v12/configuring-npm/package-json#bundledependencies) (or its `bundledDependencies` alias) are exempt, since their contents are published within the package's own tarball and so remain resolvable when installed from a registry.
+Setting `bundleDependencies` to `true` bundles every dependency, and likewise exempts them all.
+
+If both spellings are present, only `bundleDependencies` is honored, matching how npm resolves the alias.
```
michaelfaith: Based on the behavior, should this be "both spellings are present _and truthy_"?  Currently if `bundleDependencies` is present and `false`, it's ignored.  Is that consistent with how npm handles the fallback?

src/tests/rules/no-local-dependencies.test.ts
```diff
@@ -135,6 +210,42 @@ ruleTester.run('no-local-dependencies', rule, {
...
+	},
+	"bundledDependencies": ["abc"]
+}`,
+    `{
+	"dependencies": {
+		"abc": "file:abc.tgz"
+	},
+	"bundleDependencies": false,
+	"bundledDependencies": ["abc"]
+}`,
```
michaelfaith, line 251: Could we include an additional test that has an unexpected value for `bundleDependencies`?

e.g. 
```json
{
	"dependencies": {
		"abc": "file:abc.tgz"
	},
	"bundleDependencies": 123,
	"bundledDependencies": ["abc"]
}
```

src/rules/no-local-dependencies.ts
```diff
@@ -27,19 +56,54 @@ export const rule = createRule({
...
 
+        let bundleDependenciesValue =
+          bundleDependencyValues.get('bundleDependencies');
+        // npm only falls back to the `bundledDependencies` spelling when
+        // `bundleDependencies` is absent or falsy, so the two never combine.
+        if (!(
+          bundleDependenciesValue &&
+          (bundleDependenciesValue.type !== 'JSONLiteral' ||
+            Boolean(bundleDependenciesValue.value))
+        )) {
```
michaelfaith, line 82: I suspect this might break if `bundleDependencies` is a literal but not a boolean.  That's not a valid value, but we wouldn't want this rule to break because of an invalid value.  I suggested another test to include such a case.  Unless that's how npm would also handle a truthy literal value that's not actually a boolean...?

michaelfaith (commented):
As always, great work!  Thanks for the quick action here.  I left a couple bits of feedback and highlighted a possible bug.  Beyond those pieces of feedback, this looks spot on 👍

michaelfaith:
@andreww2012 hey there 👋. just wanted to see if you needed a hand at all on the remainder of the changes?

andreww2012:
Hey, I have been on vacation, I should be able to address the comments tomorrow or in the coming days 🙂

src/rules/no-local-dependencies.ts
andreww2012: Added a new test confirming the rule still works correctly even if `bundleDependencies` is a non-boolean literal

site/src/content/docs/rules/no-local-dependencies.mdx
andreww2012: Clarified the behavior better. And yes, [it is the npm behavior](https://github.com/npm/normalize-package-data/blob/34c98503c4e828a166aa3bb78a4cb0525af3c8f5/lib/fixer.js#L113)

michaelfaith (approved):
Great work, as always!  Thanks again for the support 🙏

PR #2054, opened 2026-07-31
feat(restrict-dependency-ranges): add `<`, `<=`, `>`, `>=` as possible `rangeTypes`

### LINKED ISSUE #2028: 🚀 Feature: `restrict-dependency-ranges` - add options to support restrictions involving `>`/`>=`/`<`/`<=`
### Feature Request Checklist

### Overview

it would be nice if the config for `restrict-dependency-ranges` supported a `rangeType` that enforced only the following symbols: `>`, `>=`, `<`, `<=`

### Additional Info

in my project, we don't need to worry about pinning dependencies in `package.json` because the dependencies will always be installed using the locked versions from `package-lock.json`. this is because my project is a bundled application, not a published npm package.

this makes it much easier for us to bump our dependencies because we just run `npm update` to get the latest versions of all our dependencies, and if we encounter any breaking changes that we can't resolve, only then do we add an upper bound to that dependency in our  `package.json` (using `<` or `<=`)

DetachHead:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2028 


DetachHead:
> [[**pr-branch-non-default**](https://octo.guide/rules/pr-branch-non-default)] This PR is sent from the head repository's default branch Sending a PR from a default branch means the head repository can't easily be updated after the PR is merged. Sending a PR from a repository's default branch, commonly `main`, means that repository will have a hard time pulling in updates from the upstream repository. It's generally recommended to instead create a new branch per pull request. You'll need to:

IMO it shouldn't matter to this repo how i manage my branches in my own fork. i find it more convenient to just use the main branch and only make separate branches when i intend to make multiple contributions

michaelfaith:
> IMO it shouldn't matter to this repo how i manage my branches in my own fork.

I agree.  And that's why, after seeing this, i disabled the rule [here](https://github.com/michaelfaith/eslint-plugin-package-json/pull/2055), and filed an issue upstream to remove it from their `recommended` config [here](https://github.com/octoguide/bot/issues/624).

src/rules/restrict-dependency-ranges.ts
```diff
@@ -11,9 +11,27 @@ const DEPENDENCY_TYPES = [
...
 
+const SYMBOLS = RANGE_TYPES.filter((rangeType) => 'symbol' in rangeType).map(
+  (rangeType) => rangeType.symbol,
+);
+type SymbolType = (typeof SYMBOLS)[number];
+
+const RANGE_NAMES = RANGE_TYPES.filter((rangeType) => 'alias' in rangeType).map(
+  (rangeType) => rangeType.alias,
+);
+type RangeNames = (typeof RANGE_NAMES)[number];
```
michaelfaith: Nit

```suggestion
type RangeName = (typeof RANGE_NAMES)[number];
```

src/rules/restrict-dependency-ranges.ts
```diff
@@ -11,9 +11,27 @@ const DEPENDENCY_TYPES = [
...
+  { symbol: '<=' },
+  { symbol: '>' },
+  { symbol: '>=' },
+] as const;
 type RangeType = (typeof RANGE_TYPES)[number];
 
+const SYMBOLS = RANGE_TYPES.filter((rangeType) => 'symbol' in rangeType).map(
+  (rangeType) => rangeType.symbol,
+);
+type SymbolType = (typeof SYMBOLS)[number];
```
michaelfaith:
```suggestion
type RangeSymbol = (typeof SYMBOLS)[number];
```

src/rules/restrict-dependency-ranges.ts
```diff
@@ -11,9 +11,27 @@ const DEPENDENCY_TYPES = [
...
 
-const RANGE_TYPES = ['caret', 'pin', 'tilde'] as const;
+const RANGE_TYPES = [
+  { symbol: '^', alias: 'caret' },
+  { alias: 'pin', workspaceSymbol: '*' },
+  { symbol: '~', alias: 'tilde' },
+  { symbol: '<' },
+  { symbol: '<=' },
+  { symbol: '>' },
+  { symbol: '>=' },
```
michaelfaith: We could include aliases for these too (e.g. `lt`, `lte`, `gt`, `gte`).  Though, I also don't feel super strongly about it.  So happy to leave that off.

src/rules/restrict-dependency-ranges.ts
```diff
@@ -70,20 +116,11 @@ const changeVersionRange = (version: string, rangeType: RangeType): string => {
...
   if (/^workspace:[~^*]$/.test(version)) {
-    switch (rangeType) {
-      case 'caret':
-        return 'workspace:^';
-      case 'pin':
-        return 'workspace:*';
-      default:
-        return 'workspace:~';
-    }
+    return workspaceVersionStartsWith(rangeType);
```
michaelfaith: I don't believe `workspace:` protocol will work with only inequality symbols.  e.g. `workspace:<`, which this logic would potentially create.

src/rules/restrict-dependency-ranges.ts
```diff
@@ -153,13 +186,17 @@ export const rule = createRule({
...
           }
 
-          const isPinned = !!semver.parse(version) || version === 'workspace:*';
-          const isTildeRange =
-            (!!semver.validRange(version) && version.startsWith('~')) ||
-            version.startsWith('workspace:~');
-          const isCaretRange =
-            (!!semver.validRange(version) && version.startsWith('^')) ||
-            version.startsWith('workspace:^');
+          const isRangeType = (rangeType: RangeType) => {
```
michaelfaith:
```suggestion
          const doesRangeTypeMatch = (rangeType: RangeType) => {
```

src/rules/restrict-dependency-ranges.ts
```diff
@@ -196,37 +233,30 @@ export const rule = createRule({
...
-              switch (rangeType) {
-                case 'caret':
-                  return isCaretRange;
-                case 'pin':
-                  return isPinned;
-                case 'tilde':
-                  return isTildeRange;
-              }
-            });
+            const rangeTypeMatch = rangeTypes.some(isRangeType);
```
michaelfaith:
```suggestion
            const isRangeTypeValid = rangeTypes.some(isRangeType);
```

src/rules/restrict-dependency-ranges.ts
```diff
@@ -259,9 +292,7 @@ export const rule = createRule({
     },
     hasSuggestions: true,
     messages: {
-      changeToCaret: 'Change to use a caret range.',
-      changeToPin: 'Pin the version.',
-      changeToTilde: 'Change to use a tilde range.',
+      changeTo: 'Change to use a {{rangeType}}',
```
michaelfaith:
```suggestion
      changeTo: 'Change to use a {{rangeType}} range',
```

And i'm ok with this consolidation with all of them except `pin`.  I'd like to keep the original `pin` wording.  "Change to use a pin" doesn't read very well.

michaelfaith (changes requested):
Great start.  Thanks again for taking this up.  I've left a bit of feedback.  Also, the doc page for this rule needs to be updated too, to reflect the new options.  And it would be good to add another test to cover the missing line/condition.

src/rules/restrict-dependency-ranges.ts
DetachHead: i'm not really familiar with `workspace:` dependencies. do you mean that `<`, `<=`, `>` and `>=` are all not supported? and that we should only call `workspaceVersionStartsWith` if the range type is `*`, `~` or `"pin"`?

src/rules/restrict-dependency-ranges.ts
```diff
@@ -57,11 +82,30 @@ const schemaOptions = {
...
+const normalizeRangeType = (
+  rangeTypeOrSymbol: RangeName | RangeSymbol,
+): RangeType => {
+  const result = RANGE_TYPES.find(
+    (rangeType) =>
+      ('symbol' in rangeType && rangeType.symbol === rangeTypeOrSymbol) ||
+      rangeType.alias === rangeTypeOrSymbol,
+  );
+  if (result === undefined) {
+    throw new Error(`invalid range type: ${rangeTypeOrSymbol}`);
```
DetachHead: >  And it would be good to add another test to cover the missing line/condition.

hmm i think this line is unreachable, since specifying an invalid range type causes the json schema validation to fail before this line is reached:

```
Error: Key "rules": Key "package-json/restrict-dependency-ranges":
        Value ["asdfasdfafd"] should be equal to one of the allowed values.
        Value "asdfasdfafd" should be equal to one of the allowed values.
        Value ["asdfasdfafd"] should be equal to one of the allowed values.
        Value "asdfasdfafd" should be equal to one of the allowed values.
        Value ["asdfasdfafd"] should match exactly one schema in oneOf.
        Value {"rangeType":["asdfasdfafd"]} should be array.
        Value {"rangeType":["asdfasdfafd"]} should match exactly one schema in oneOf.
```
the only reason i added this line was to avoid a typescrtipt error. i also tried using `!` instead but there's an eslint rule banning it.

i added a `codecov:ignore` comment, not sure if that's the best solution tho

src/rules/restrict-dependency-ranges.ts
```diff
@@ -57,11 +82,31 @@ const schemaOptions = {
...
+  );
+  if (result === undefined) {
+    // codecov:ignore-next-line - this should be unreachable due to the JSON schema validation
+    throw new Error(`invalid range type: ${rangeTypeOrSymbol}`);
+  }
+  return result;
+};
+
+const workspaceVersionStartsWith = (rangeType: RangeType) => {
+  if ('workspaceSymbol' in rangeType) {
```
michaelfaith, line 90: This function name isn't very reflective of its purpose.  How about `getWorkspaceVersionForRange`? (sorry for the delay on this, I'd typed it out last review, but didn't hit submit 🤦)

src/rules/restrict-dependency-ranges.ts
```diff
@@ -42,11 +58,20 @@ const schemaOptions = {
...
         {
-          enum: RANGE_TYPES,
+          enum: SYMBOLS,
+        },
+        {
+          items: {
+            enum: SYMBOLS,
+          },
+          type: 'array',
+        },
```
michaelfaith: The way this is setup, does it mean that we can't have an array that includes both symbols and names?  I think the config should be interchangeable.  i.e. `["<", "tilde", "gte"]` should work (and be equivalent to `["lt", "~", ">="]`. Since `pin` only exists as a name, it's necessary to be able to support a mix.  Would be good to include a test case for this, to ensure we do.

src/rules/restrict-dependency-ranges.ts
michaelfaith: If the code is unreachable, then I'd rather not have it, and go with the `!` + `eslint-disable`

src/rules/restrict-dependency-ranges.ts
michaelfaith: > and that we should only call `workspaceVersionStartsWith` if the range type is *, ~ or "pin"?

Basically.  `workspace:>` is not support, but `workspace:>2.0.0` is supported.  Whereas `workspace:^`, `workspace:~`, and `workspace:*`  are also supported.  So we wouldn't want to swap `workspace:^` for `workspace:>`, since that's not valid, but we would want to swap `workspace:^1.0.0` with `workspace:>1.0.0`

michaelfaith: What I'm undecided on, though, is if we should even report those cases... 🤔  I'm inclined to leave those alone, since they have special meaning, but I could see both sides.

michaelfaith (changes requested):
Good progress.  Answered a couple of questions, and pointed out a potential bug in the schema setup.

src/rules/restrict-dependency-ranges.ts
DetachHead: ok i've updated it to only report on `rangeType`s that support `workspace:` (i think. sorry i still don't fully understand how these work)

src/rules/restrict-dependency-ranges.ts
DetachHead: fixed. however that raises a question: what to do if the user specifies the same `rangeType` twice eg. `['tilde', '~']` should it throw an error? i would say yes, but perhaps that's outside the scope of this PR since i think `['tilde', 'tilde']` is currently allowed too

src/rules/restrict-dependency-ranges.ts
```diff
@@ -57,50 +75,61 @@ const schemaOptions = {
...
+    return `workspace:${rangeType.workspaceSymbol}`;
+  }
+};
+
+/** For displaying a range type in a user-facing way (ie. an error message). */
+const displayRangeType = (rangeType: RangeType) =>
+  'symbol' in rangeType ? rangeType.symbol : rangeType.alias;
+
+const isWorkspaceSymbol = (version: string) =>
+  /^workspace:[~^*]$/.test(version);
```
michaelfaith:
```suggestion
const isRollingWorkspaceSpec = (version: string) =>
  /^workspace:[~^*]$/.test(version);
```

Also, could we extract out the Regex's to constants outside of the functions.  I know it wasn't like that before for the ones that were already there, but now that we're adding more, it would be nice to do that.  So we're not re-initializing the same regex's over and over again with every function call.

src/rules/restrict-dependency-ranges.ts
```diff
@@ -57,50 +75,61 @@ const schemaOptions = {
...
+
+const isWorkspaceSymbolWithVersion = (version: string) =>
+  /^workspace:[~^*]/.test(version);
```
michaelfaith: Nit on the naming (this function name doesn't really reflect what it's testing.  This not only detects `workspace:^1.2.3` but also `workspace:^`.  It's named as if it only detects the first.  Coupled with feedback below, should we just be testing if it's workspace protocol at this point?  e.g. `startsWith('workspace:')`  If so, then maybe it's just `isWorkspaceProtocol`...

src/rules/restrict-dependency-ranges.ts
```diff
@@ -153,13 +182,25 @@ export const rule = createRule({
...
+            if ('symbol' in rangeType && semver.validRange(version)) {
+              return version.startsWith(rangeType.symbol);
+            }
+            if (
+              rangeType.alias === 'pin' &&
+              (!!semver.parse(version) || version === 'workspace:*')
+            ) {
+              return true;
+            }
+            if (isWorkspaceSymbolWithVersion(version)) {
```
michaelfaith: I think there's still a gap here.  This function only catches changes to `workspace:[~^*]` but doesn't attempt to change `workspace:>=1.2.3`.  If the expected range type is `^` that should be changed to `workspace:^1.2.3`, right?  I don't believe that will be supported here. (it would be good to add tests for that)

 This filter function also isn't exhaustive.  What if none of the conditions are met?  It just returns `undefined`.  Does that mean the range type doesn't match?  Or we don't know if it matches?

src/rules/restrict-dependency-ranges.ts
michaelfaith: Mmm, good callout.  I think adding `uniqueItems: false` should fix the existing repeat issue, but wouldn't solve the new one. I wouldn't want to throw an error, so I think it's ok to leave it for now.

michaelfaith (changes requested):
Good progress!  Still a few more areas of feedback.  Almost there.

src/rules/restrict-dependency-ranges.ts
DetachHead: done, though the regexes are slightly different so i had to go ``new RegExp(`${WORKSPACE_REGEX.source}$`)`` which imo is more difficult to read

DetachHead: nvm, i just removed the other function based on your other comments and inlined it again since it still only has one usage

src/rules/restrict-dependency-ranges.ts
DetachHead: removed this function and replaced it with an inline `startsWith('workspace:')` check

src/rules/restrict-dependency-ranges.ts
DetachHead: > I think there's still a gap here. This function only catches changes to `workspace:[~^*]` but doesn't attempt to change `workspace:>=1.2.3`. If the expected range type is `^` that should be changed to `workspace:^1.2.3`, right? I don't believe that will be supported here. (it would be good to add tests for that)

i added a test for this and it seems to work as expected (https://github.com/michaelfaith/eslint-plugin-package-json/pull/2054/changes/7bd13eb413bc79953fa56d86ab7f25e165baaa5c)

> This filter function also isn't exhaustive. What if none of the conditions are met? It just returns `undefined`. Does that mean the range type doesn't match? Or we don't know if it matches?

good catch. i enabled `noImplicitReturns` in `tsconfig.json` to avoid mistakes like this in the future (https://github.com/michaelfaith/eslint-plugin-package-json/pull/2054/changes/8a44b14922dd44ccb28221dd206177d22eec72e1)

src/rules/restrict-dependency-ranges.ts
```diff
@@ -153,13 +180,26 @@ export const rule = createRule({
...
-          const isCaretRange =
-            (!!semver.validRange(version) && version.startsWith('^')) ||
-            version.startsWith('workspace:^');
+          const doesRangeTypeMatch = (rangeType: RangeType) => {
+            if ('symbol' in rangeType && semver.validRange(version)) {
+              return version.startsWith(rangeType.symbol);
+            }
+            if (
+              rangeType.alias === 'pin' &&
+              (!!semver.parse(version) || version === 'workspace:*')
```
michaelfaith:
```suggestion
              (!!semver.parse(version) || version === 'workspace:*' || version === `workspace:`)
```

`workspace:` (with no qualifier) behaves the same as `workspace:*`

src/rules/restrict-dependency-ranges.ts
```diff
@@ -57,50 +75,59 @@ const schemaOptions = {
...
+  }
+  return undefined;
+};
+
+/** For displaying a range type in a user-facing way (ie. an error message). */
+const displayRangeType = (rangeType: RangeType) =>
+  'symbol' in rangeType ? rangeType.symbol : rangeType.alias;
+
+const isRollingWorkspaceSpec = (version: string) =>
+  /^workspace:[~^*]$/.test(version);
```
michaelfaith: So, this regex still needs to be pulled out of the function (as well as the one on 117).  It's not about reuse across callsites.  It's about runtime instantiations.  Every time this function is called, it's going to `new` up a `RegExp` object, even though the regex pattern is the same.  If you pull the regex out as a const in the global scope, it only creates one instance.

michaelfaith (changes requested):
Looks good; just two minor tweaks, and should be good to go.  Thanks!

src/rules/restrict-dependency-ranges.ts
DetachHead: done. there were other regexes that i extracted too (but i couldn't really think of good names for them)

michaelfaith (approved):
Looks good 👍  Thanks again for the contribution.

michaelfaith:
@all-contributors please add @DetachHead for code

PR #2066, opened 2026-08-04
fix(scripts-name-casing): support "hidden" scripts and `pnpm:devPreinstall`

### LINKED ISSUE #2047: 🐛 Bug: `scripts-name-casing` should ignore `pnpm:devPreinstall` (and maybe also hidden pnpm scripts starting with `.`)
### Bug Report Checklist

### Expected

Similar to #1368: `pnpm:devPreinstall` is the special [script only supported by pnpm](https://pnpm.io/scripts#pnpmdevpreinstall).

Additionally, starting with v11, pnpm also supports so called ["hidden scripts" which names start with a dot](https://pnpm.io/scripts#hidden-scripts).

Should we ignore such script names in regards to this rule?

I also kind of want to raise the not being able to ignore script names problem again, since who knows what other conventions package managers (i.e. possibly more niche ones like the recently born [nub](https://nubjs.com)) might come up with.

### Actual

The following script names get flagged by the reported rule and it's not possible to suppress that:

```json
{
  "pnpm:devPreinstall": "echo foo",
  ".bar": "echo bar"
}
```

### Version

1.6.2

### ESLint Version

10.8.0

### Additional Info

(just a sketch; happy to create a separate issue if we agree on the need of this option)

### Proposed `ignore` option

Should accept an array of strings. If a string starts with `/` and ends with `/<optional regex flag(s)>`, it's treated as a regex; otherwise as a plain string (on the edge case: this would be super weird to ignore a script name starting literally with `/`, but that would still be possible to ignore using the regex form).

The example of configuring the rule to ignore the reported script names:

```ts
const rules = {
  'package-json/scripts-name-casing': [
    2,
    {
      ignore: ['pnpm:devPreinstall', '/^\\.[a-z0-9]+(-[a-z0-9]+)*$/'],
    },
  ],
};
```

michaelfaith:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2047

## Overview

This change adds support for both the [built-in](https://pnpm.io/scripts#pnpmdevpreinstall) `pnpm:devPreinstall` and pnpm's ["hidden" scripts](https://pnpm.io/scripts#hidden-scripts).

For hidden scripts, we're still validating that the hidden script is kebab-case, we're just ignoring the leading `.`, when performing validation and making fix suggestions.

### DIFF
--- a/src/rules/scripts-name-casing.ts
+++ b/src/rules/scripts-name-casing.ts
@@ -4,7 +4,10 @@ import type { AST } from 'jsonc-eslint-parser';
 import { createRule } from '../createRule.ts';
 
 // See https://docs.npmjs.com/cli/v11/using-npm/scripts
-const BUILT_IN_SCRIPTS_IN_CAMEL_CASE = new Set(['prepublishOnly']);
+const BUILT_IN_SCRIPTS_IN_CAMEL_CASE = new Set([
+  'prepublishOnly',
+  'pnpm:devPreinstall',
+]);
 
 export const rule = createRule({
   create(context) {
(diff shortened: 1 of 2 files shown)


michaelfaith:
cc @andreww2012

PR #2178, opened 2026-09-12
fix: do not rewrite null exports in `exports-subpaths-style`

### LINKED ISSUE #2177: 🐛 Bug: exports-subpaths-style rewrites null exports and changes Node resolution semantics
### Bug Report Checklist

### Expected

`null` and `{ ".": null }` are distinct Node.js resolution semantics and neither should be rewritten by `exports-subpaths-style`.

- `"exports": null` is treated as if the field were absent (legacy `main`/deep-file resolution).
- `"exports": { ".": null }` is a non-null `exports` object, so the package is encapsulated and the root target resolves to nothing: `ERR_PACKAGE_PATH_NOT_EXPORTED`.

### Actual

The rule treats the two as interchangeable single-root-export formats and autofixes between them, silently changing runtime resolution.

Default (`prefer: "explicit"`):

```jsonc
// input
{ "exports": null }
// reported, fixed to
{ "exports": { ".": null } }
```

`prefer: "implicit"`:

```jsonc
// input
{ "exports": { ".": null } }
// reported, fixed to
{ "exports": null }
```

### Version

1.8.0

### ESLint Version

10.10.0 (also reproduces on 9.x; the behavior is independent of the ESLint major)

### Additional Info

Node.js resolves these differently. From Node v24.13.1:

- A top-level `null` skips exports resolution entirely. ESM gates on `if (packageConfig.exports != null)` (`lib/internal/modules/esm/resolve.js#L774`) and otherwise falls back to `legacyMainResolve`/URL resolution. CJS mirrors it: `if (pkg.exists && pkg.exports != null)` (`lib/internal/modules/cjs/loader.js#L676`).
- `{ ".": null }` is a non-null object, so Node takes the exports path (`#L774`), and a null target resolves to nothing: `if (resolveResult == null) { throw exportsNotFound(...) }` (`resolve.js#L588`, `#L604`) -> `ERR_PACKAGE_PATH_NOT_EXPORTED`.
- Docs: `doc/api/packages.md#L370` ("When the `"exports"` field is defined, all subpaths of the package are encapsulated...") and `#L636` ("To exclude private subfolders from patterns, `null` targets can be used.").

The current test suite encodes the bug: `{ "exports": null }` is in the `invalid` list with output `{ "exports": { ".": null } }`, and there is no implicit-mode case for `{ ".": null }`.

Suggested fix: skip JSON `null` literals in both `validateForExplicit` and `validateForImplicit`. Non-null targets are unaffected (`{ ".": "./x" }` <-> `"./x"` remain equivalent). I have a failing-test-first PR ready.

privatenumber:
## Problem

`exports-subpaths-style` treats `null` and `{ ".": null }` as equivalent single-root-export formats and autofixes between them. Node.js resolves them differently, so the fix silently changes runtime behavior.

A package sets `"exports": { ".": null }` to encapsulate itself and expose no entry points. With `prefer: "implicit"`, `--fix` rewrites it to `"exports": null`; Node treats a top-level `null` as if the field were absent, so legacy `main` and deep-file imports start resolving again. The default `prefer: "explicit"` mode corrupts the reverse case.

The difference is in Node's resolvers (v24.13.1): a top-level `null` is gated out by [`if (packageConfig.exports != null)`](https://github.com/nodejs/node/blob/v24.13.1/lib/internal/modules/esm/resolve.js#L774) (CJS mirrors it in [`loader.js#L676`](https://github.com/nodejs/node/blob/v24.13.1/lib/internal/modules/cjs/loader.js#L676)) and falls back to legacy resolution. `{ ".": null }` is a non-null object, so the exports path runs and a null target [throws `ERR_PACKAGE_PATH_NOT_EXPORTED`](https://github.com/nodejs/node/blob/v24.13.1/lib/internal/modules/esm/resolve.js#L604). The docs describe [encapsulation when `exports` is defined](https://github.com/nodejs/node/blob/v24.13.1/doc/api/packages.md#L370) and [null targets excluding subpaths](https://github.com/nodejs/node/blob/v24.13.1/doc/api/packages.md#L636).

Closes #2177

## Changes

`null` literal targets are now skipped in both `validateForExplicit` and `validateForImplicit`, so neither mode rewrites them. Non-null targets are unaffected: `{ ".": "./x" }` and `"./x"` still normalize as before.

The obsolete `{ "exports": null }` invalid case moves to `valid`, and implicit-mode coverage for `{ ".": null }` is added.

### DIFF
--- a/src/rules/exports-subpaths-style.ts
+++ b/src/rules/exports-subpaths-style.ts
@@ -1,7 +1,10 @@
 import type { AST } from 'jsonc-eslint-parser';
 
 import { createRule } from '../createRule.ts';
-import { isJSONStringLiteral } from '../utils/predicates/index.ts';
+import {
+  isJSONNullLiteral,
+  isJSONStringLiteral,
+} from '../utils/predicates/index.ts';
 
 function isImplicitFormat(
   node: AST.JSONLiteral | AST.JSONObjectExpression,
@@ -24,6 +27,14 @@ export const rule = createRule({
 
     function validateForExplicit(node: AST.JSONProperty) {
       const { value } = node;
+
+      // A top-level null is not a root export; Node treats it as if the
+      // exports field were absent (legacy resolution). Wrapping it would
+      // enable encapsulation and change runtime behavior.
+      if (isJSONNullLiteral(value)) {
+        return;
+      }
+
       if (
         (value.type !== 'JSONLiteral' &&
           value.type !== 'JSONObjectExpression') ||
@@ -64,6 +75,13 @@ export const rule = createRule({
       }
 
       const dotProperty = value.properties[0];
+
+      // A null target means "not exported". Rewriting `{ ".": null }` to
+      // `null` removes encapsulation and restores legacy resolution.
+      if (isJSONNullLiteral(dotProperty.value)) {
+        return;
+      }
+
       context.report({
         fix(fixer) {
           const valueText = context.sourceCode.getText(dotProperty.value);
--- a/src/tests/rules/exports-subpaths-style.test.ts
+++ b/src/tests/rules/exports-subpaths-style.test.ts
@@ -44,13 +44,6 @@ ruleTester.run('exports-subpaths-style', rule, {
   ".": true
 } }`,
     },
-    {
-      code: `{ "exports": null }`,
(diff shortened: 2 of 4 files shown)


michaelfaith (approved):
Great work, as always. 👏 Thanks again!

PR #2189, opened 2026-09-15
feat: add `restrict-dist-tags` rule (#2169)

### LINKED ISSUE #2169: 🚀 Feature: add new `restrict-dist-tags` rule
### Feature Request Checklist

### Overview

Create a new rule named `restrict-dist-tags` that puts similar restrictions as `restrict-dependency-ranges` does, but to specs defined as `dist-tags` (e.g. `beta`).  This api is a bit simpler than `restrict-dependency-ranges` though.

#### API
The options type should be 

```ts
interface Options {
  allowed: string[];
  allowedFor: 'dependencies' | 'devDependencies' | 'peerDependencies' | 'optionalDependencies';
}
```

#### Examples

User wants to only allow dist-tags for dev dependencies, and prevent them for peer and regular deps

```ts
export default [
  {
    'package-json/restrict-dist-tags': [
      'error',
      {
        allowedFor: ['devDependencies'],
      },
    ],
  },
];
```

User wants to only allow dist-tags for dev dependencies, and prevent them for peer and regular deps.  And restrict the dist-tags allowed to only `beta` or higher.

```ts
export default [
  {
    'package-json/restrict-dist-tags': [
      'error',
      {
        allowedFor: ['devDependencies'],
        allowed: ['beta', 'rc'],
      },
    ],
  },
];
```

User only wants to allow `rc` pre-releases and doesn't care which dependency type they're in.

```ts
export default [
  {
    'package-json/restrict-dist-tags': [
      'error',
      {
        allowed: ['rc'],
      },
    ],
  },
];
```

ShayanTheNerd:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2169

## Overview

I Implemented `restrict-dist-tags` for restricting [npm dist-tags](https://docs.npmjs.com/cli/v8/commands/npm-dist-tag) used in dependency declarations.

The rule takes `allowed` and `allowedFor` as options and limits the permitted dist-tags and dependency types. An omitted `allowed` permits all common dist-tags. An omitted or empty `allowedFor` applies the allow-list to all dependency types. Dist-tags used outside the selected dependency types are reported, even when the tag itself is allowed.

The rule ignores semver versions and ranges, workspace specifications, local paths, Git specifications, URLs, npm aliases, catalog specifications, and non-string dependency values.

I also added comprehensive RuleTester coverage and documentation for required options, empty allow-lists, dependency-type scoping, and ignored dependency specifications.

✨


src/tests/rules/restrict-dist-tags.test.ts
```diff
@@ -0,0 +1,254 @@
...
+        {
+          data: {
+            dependencyType: 'devDependencies',
+            tag: 'alpha',
+          },
+          messageId: 'disallowedDistTag',
+        },
+      ],
+      name: 'empty allowedFor applies allowed tags to every dependency type',
+      options: [{ allowed: ['rc'], allowedFor: [] }],
```
michaelfaith: There's a bit of asymmetry here, in terms of how we treat empty arrays, and I'm thinking that could be confusing to the users.  For `allowed` and empty array means _deny_ all dist-tags.  For `allowedFor`, and empty array means _allow_ all dependency types.  Should we make empty array behave the same for each?

src/rules/restrict-dist-tags.ts
```diff
@@ -0,0 +1,128 @@
...
+import { isJSONStringLiteral } from '../utils/predicates/index.ts';
+
+const DEPENDENCY_TYPES = [
+  'dependencies',
+  'devDependencies',
+  'optionalDependencies',
+  'peerDependencies',
+] as const;
+
+type Dependency = (typeof DEPENDENCY_TYPES)[number];
```
michaelfaith: I'm not sure I see the value of this type or by defining `DEPENDENCY_TYPES` as `as const`.  The only place the type is used it to force strings that may not be one of these four to be used in an `includes` condition.  If you remove the `as const`, you won't need to `as` cast lower down in those comparisons, and the type won't be necessary.  It's only serving a purpose because of how narrow the type of the array is.

src/rules/restrict-dist-tags.ts
```diff
@@ -0,0 +1,128 @@
...
+  'dev',
+  'next',
+  'alpha',
+  'beta',
+  'canary',
+  'rc',
+];
+
+function isDistTag(value: string): boolean {
+  return !semver.validRange(value) && /^[A-Z0-9][\w.-]*$/i.test(value);
```
michaelfaith: Recommend pulling this regex out into a constant defined at the module level, so it's not being recreated every time the function is called.

src/rules/restrict-dist-tags.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+const schema = {
```
michaelfaith: I think this can just be defined inline lower down where the variable is being used.  I don't see much value in pulling it out as a module level constant.

src/rules/restrict-dist-tags.ts
```diff
@@ -0,0 +1,128 @@
...
+      'Program > JSONExpressionStatement > JSONObjectExpression > JSONProperty[key.type=JSONLiteral][value.type=JSONObjectExpression]'(
+        node: AST.JSONProperty & {
+          key: AST.JSONStringLiteral;
+          value: AST.JSONObjectExpression;
+        },
+      ) {
+        const dependencyType = node.key.value;
+
+        if (!DEPENDENCY_TYPES.includes(dependencyType as Dependency)) {
+          return;
```
michaelfaith, line 44: This condition isn't covered in the unit tests currently.  If you were to add a test to `valid` with a key that isn't included in `DEPENDENCY_TYPES`, it should cover it.

src/rules/restrict-dist-tags.ts
```diff
@@ -0,0 +1,128 @@
...
+    },
+  },
+  additionalProperties: false,
+} as const;
+
+const rule = createRule({
+  create(context) {
+    const options = context.options[0] ?? {};
+    const allowedTags = options.allowed ?? DEFAULT_ALLOWED_TAGS;
+    const allowedForDependencies = options.allowedFor ?? [];
```
michaelfaith: Nit: this can be simplified
```suggestion
    const { allowed = DEFAULT_ALLOWED_TAGS, allowedFor = [] } = context.options[0] ?? {};
```

src/rules/restrict-dist-tags.ts
```diff
@@ -0,0 +1,128 @@
...
+        }
+      },
+    };
+  },
+
+  meta: {
+    defaultOptions: [
+      {
+        allowed: DEFAULT_ALLOWED_TAGS,
+        allowedFor: [],
```
michaelfaith: Why do we need a default for `allowedFor`?  Couldn't `undefined` represent "no opinion" and allow for all dependency types?

src/rules/restrict-dist-tags.ts
```diff
@@ -0,0 +1,128 @@
...
+
+        for (const property of node.value.properties) {
+          if (
+            !isJSONStringLiteral(property.key) ||
+            !isJSONStringLiteral(property.value)
+          ) {
+            continue;
+          }
+
+          const tag = property.value.value;
```
michaelfaith: Nit: we don't know if this is a tag yet, since we haven't passed the condition below.  I think `spec` would be more accurate here.

src/rules/restrict-dist-tags.ts
```diff
@@ -0,0 +1,128 @@
...
+
+          if (!isDistTag(tag)) {
+            continue;
+          }
+
+          if (
+            (allowedForDependencies.length === 0 ||
+              allowedForDependencies.includes(dependencyType as Dependency)) &&
+            allowedTags.includes(tag)
```
michaelfaith: If you don't define a default to `allowedFor`, then this becomes


```suggestion
            allowedTags.includes(tag) && 
            (!allowedForDependencies?.length ||
              allowedForDependencies.includes(dependencyType))
            
```

site/src/content/docs/rules/restrict-dist-tags.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+- `latest`
+- `dev`
+- `next`
+- `alpha`
+- `beta`
+- `canary`
+- `rc`
+
+And other dist-tags are reported.
```
michaelfaith:
```suggestion
Any dist-tags not included in the `allowed` list are reported.
```

site/src/content/docs/rules/restrict-dist-tags.mdx
```diff
@@ -0,0 +1,147 @@
...
+
```
michaelfaith, line 51: It would be nice to include at least one example (with "incorrect" and "correct" tabs) above the `Options` section, just for the default case.

site/src/content/docs/rules/restrict-dist-tags.mdx
```diff
@@ -0,0 +1,147 @@
...
+      "devDependencies": {
+        "package-a": "canary",
+        "package-b": "rc"
+      }
+    }
+    ```
+  </TabItem>
+</Tabs>
+
+This rule only checks bare dist-tag values. It ignores semver versions and ranges, workspace specifications, local paths, Git specifications, URLs, npm aliases, catalog specifications, and non-string dependency values.
```
michaelfaith: This is great to include, but it feels out of place being at the bottom under Options.  Would you mind moving this just above Options in an `Aside` component with type "note"?  You can see an example of that in `prefer-rolling-workspace-spec`

michaelfaith (changes requested):
Thanks for the contribution!  And great start so far. 👏  Just a handful of relatively minor feedback.

ShayanTheNerd:
Sure! I just modified the implementation based on the requested changes, and updated the docs/tests accordingly.

site/src/content/docs/rules/restrict-dist-tags.mdx
```diff
@@ -0,0 +1,174 @@
...
+    ```
+  </TabItem>
+</Tabs>
+
+## Options
+
+| Name         | Type                                                                                      | Default                        |
+| ------------ | ----------------------------------------------------------------------------------------- | ------------------------------ |
+| `allowed`    | `string[]`                                                                                | Common npm dist-tags           |
+| `allowedFor` | `('dependencies' \| 'devDependencies' \| 'optionalDependencies' \| 'peerDependencies')[]` | omitted (all dependency types) |
```
michaelfaith:
```suggestion
| `allowedFor` | `('dependencies' \| 'devDependencies' \| 'optionalDependencies' \| 'peerDependencies')[]` | `undefined` (all dependency types) |
```

michaelfaith (approved):
Looks great.  Thanks again for the contribution! 👏

